PR #3, opened 2022-09-08
Aip-on-azure

violetbrina:


cpg_infra/abstraction/azure.py
```diff
@@ -2,48 +2,187 @@
...
+AZURE_BILLING_EXPIRY_DATE = "3141-25-09T00:00:00Z"
 class AzureInfra(CloudInfraBase):
     def __init__(
         self, config: CPGInfrastructureConfig, dataset_config: CPGDatasetConfig
     ):
         super().__init__(config, dataset_config)
 
-        self.resource_group_name = f'{config.dataset_storage_prefix}{self.dataset}'
-        self.storage_account_name = f'{config.dataset_storage_prefix}{self.dataset}'
+        self.region = 'australiaeast'
```
illusional, line 29: Can we make this configurable?

cpg_infra/abstraction/azure.py
```diff
@@ -2,48 +2,187 @@
...
+                    start_date=AZURE_BILLING_START_DATE,
+                    end_date=AZURE_BILLING_EXPIRY_DATE
+                ),
+                filter=filters
+            )
         )
 
     def bucket_rule_undelete(self, days=UNDELETE_PERIOD_IN_DAYS) -> Any:
-        pass
+        az.storage.BlobServiceProperties(
```
illusional, line 148: Return this?

cpg_infra/abstraction/azure.py
violetbrina: We certainly can, but the gcp implementation currently doesn't either. Should we change both?

cpg_infra/abstraction/azure.py
illusional: This needs a comment on why it's not being returned - or the method needs to be renamed? What happens if you call it twice with different days of the week - is it referenced specifically later?

cpg_infra/abstraction/azure.py
```diff
@@ -22,28 +29,200 @@ def __init__(
...
+            blob_services_name='default',
+            delete_retention_policy=az.storage.DeleteRetentionPolicyArgs(
+                days=days, enabled=True
+            ),
+            resource_group_name=self.resource_group.name,
         )
 
     def bucket_rule_undelete(self, days=UNDELETE_PERIOD_IN_DAYS) -> Any:
-        pass
+        self.storage_account_undelete_rule = True
```
illusional, line 190: Need a follow-up comment here.

cpg_infra/abstraction/azure.py
```diff
@@ -54,43 +233,73 @@ def create_bucket(
...
         return az.storage.BlobContainer(
-            f'bucket-{name}',
-            resource_group_name=self.resource_group_name,
-            account_name=self.storage_account_name,
+            f'{name}',
+            account_name=self.storage_account.name,
+            resource_group_name=project or self.resource_group.name,
             container_name=name,
+            metadata={'bucket': name},
+            public_access=az.storage.PublicAccess.BLOB,
```
illusional, line 263: Can we turn off public access?

cpg_infra/abstraction/azure.py
```diff
@@ -54,43 +233,73 @@ def create_bucket(
...
         self, resource_key: str, bucket, member, membership
     ) -> Any:
-        az.authorization.RoleAssignment(
-            resource_key,
-            scope=bucket.id,
-            principal_id=member.id,
-            role_definition_id='Contributor',
-            role_assignment_name='Storage Blob Data Contributor',
-        )
+        pass
```
illusional, line 270: Any luck with this?

illusional (commented):
Looking really great! One (very annoying) thing that I hadn't realised, is that if we have pulumi resource names that are the same across different infrastructures, Pulumi will probably be confused about which resource it should create.

It's very annoying, but could you add an `azure-` prefix to all of the resources that this creates?

stack/__main__.py
```diff
@@ -18,12 +18,18 @@ def from_pulumi():
...
     config = CPGInfrastructureConfig.from_dict(get_config())
-
     pconfig = pulumi.Config()
     dataset_config = CPGDatasetConfig.from_pulumi(pconfig, dataset=pulumi.get_stack())
-
     CpgDatasetInfrastructure.deploy_all_from_config(config, dataset_config)
 
 
 if __name__ == '__main__':
+    # import debugpy
```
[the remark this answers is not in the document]
violetbrina: Sure thing

stack/cpg.toml
```diff
@@ -1,6 +1,6 @@
 [infrastructure]
 domain = "populationgenomics.org.au"
-dataset_storage_prefix = "cpg-"
+dataset_storage_prefix = "cpg"
```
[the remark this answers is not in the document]
violetbrina: Done!

cpg_infra/abstraction/azure.py
```diff
@@ -22,28 +29,200 @@ def __init__(
     ):
         super().__init__(config, dataset_config)
 
-        self.resource_group_name = f'{config.dataset_storage_prefix}{self.dataset}'
-        self.storage_account_name = f'{config.dataset_storage_prefix}{self.dataset}'
+        self.region = 'australiaeast'
```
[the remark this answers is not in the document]
violetbrina: In the cpg.toml or for each data set?

cpg_infra/abstraction/azure.py
```diff
@@ -22,28 +29,200 @@ def __init__(
...
+            resource_group_name=self.resource_group.name,
+            policy=az.storage.ManagementPolicySchemaArgs(
+                rules=self.storage_account_lifecycle_rules,
+            ),
+            management_policy_name='default',
+        )
+
+    def create_project(self, name):
+        # TODO: re-work creating shared projects in Azure
+        return az.resources.ResourceGroup(name)
```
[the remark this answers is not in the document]
violetbrina: This function gets called, so raising an error will cause the stack to fail. Creating the resource group works fine, we're just not sure if that is the final implemenation that we want for shared groups. We'll re-word the TODO instead.

cpg_infra/abstraction/azure.py
violetbrina: Done

cpg_infra/abstraction/azure.py
violetbrina: Done

cpg_infra/abstraction/azure.py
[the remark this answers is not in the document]
violetbrina: It's also hard coded in gcp.py. Should we change that too?

cpg_infra/abstraction/azure.py
violetbrina: Yup, working now!

cpg_infra/abstraction/azure.py
```diff
@@ -54,43 +246,88 @@ def create_bucket(
...
+            account_name=self.storage_account.name,
+            resource_group_name=project or self.resource_group.name,
             container_name=name,
+            metadata={'bucket': name},
+            # TODO: work out requester_pays in Azure
         )
 
+    def bucket_membership_to_role(self, membership: BucketMembership):
+        if membership in (BucketMembership.MUTATE, BucketMembership.APPEND):
+            return f'{self.subscription}/providers/Microsoft.Authorization/roleDefinitions/ba92f5b4-2d11-453d-a403-e96b0029c9fe'
```
illusional, line 281: Where does the last guid come from, could we make it a constant or add a comment?

illusional (commented):
Looking really great! I've left a few comments, and might be worth a merge with main.
When you're very close to have it ready, we should try to deploy fewgenomes across GCP and Azure, and see what issues / clashes we see :)

illusional (approved):
Awesome 🎉

stack/new_stack.py
```diff
@@ -574,32 +688,15 @@ def create_stack(
...
-        )
-    else:
-        logging.info(
-            f"""
+    logging.info(
+        f"""
 Created stack {dataset}, you can commit and push this with:
 
     git checkout -b add-{dataset}-stack
     git add {' '.join(files_to_add)}
```
vladsavelyev, line 696: Missing `git commit` here?

PR #13, opened 2022-10-26
Make generating tokens path-agnostic

illusional:
Noop change, just useful when I wanted to run `python tokens/main.py`

### DIFF
--- a/tokens/main.py
+++ b/tokens/main.py
@@ -3,12 +3,12 @@
 
 import base64
 import json
+from pathlib import Path
 
 import kubernetes.client
 import kubernetes.config
 import yaml
 from google.cloud import secretmanager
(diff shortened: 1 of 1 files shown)


lgruen (approved):

tokens/main.py
```diff
@@ -3,16 +3,21 @@
 
 import base64
 import json
+
 import kubernetes.client
 import kubernetes.config
 import yaml
 from google.cloud import secretmanager
+from pathlib import Path
```
lgruen, line 11: Linter says `tokens/main.py:11:0: C0411: standard import "from pathlib import Path" should be placed before "import kubernetes.client" (wrong-import-order)`, but GitHub doesn't let me fix it as a suggestion :(.

PR #14, opened 2022-10-26
Add lifecycle rule to GCP buckets to delete incomplete multipart uploads

illusional:
Context: https://centrepopgen.slack.com/archives/C03FZL2EF24/p1666607536387039

### DIFF
--- a/cpg_infra/abstraction/gcp.py
+++ b/cpg_infra/abstraction/gcp.py
@@ -252,9 +252,11 @@ class GcpInfrastructure(CloudInfraBase):
             unique_bucket_name = (
                 f'{self.config.dataset_storage_prefix}{self.dataset}-{name}'
             )
-
-        _lifecycle_rules = list(lifecycle_rules or [])
-        _lifecycle_rules.append(self.bucket_rule_abort_incomplete_multipart_upload())
+        # duplicate the array to avoid adding the lifecycle rule to an existing list
+        _lifecycle_rules = [
+            *lifecycle_rules,
+            self.bucket_rule_abort_incomplete_multipart_upload(),
+        ]
 
         return gcp.storage.Bucket(
             unique_bucket_name,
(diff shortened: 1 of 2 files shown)


cpg_infra/abstraction/gcp.py
```diff
@@ -224,6 +225,19 @@ def bucket_rule_temporary(self, days=TMP_BUCKET_PERIOD_IN_DAYS) -> Any:
             condition=gcp.storage.BucketLifecycleRuleConditionArgs(age=days),
         )
 
+    def bucket_rule_abort_incomplete_multipart_upload(
+        self, days=BUCKET_DELETE_INCOMPLETE_UPLOAD_PERIOD_IN_DAYS
+    ):
+        """
+        Life cycle rule that deletes incomplete multipart uploads after n days
```
lgruen, line 232:
```suggestion
        Lifecycle rule that deletes incomplete multipart uploads after n days
```

cpg_infra/abstraction/gcp.py
```diff
@@ -238,14 +252,18 @@ def create_bucket(
             unique_bucket_name = (
                 f'{self.config.dataset_storage_prefix}{self.dataset}-{name}'
             )
+
+        _lifecycle_rules = list(lifecycle_rules or [])
```
lgruen, line 256: If we allow `lifecycle_rules` being `None`, should this be reflected in the type hint as `list | None`?

Do we need the cast to `list` here?

Side nit: `versioning` and `project` parameters have a space for the default parameter assignment -- not sure why the linter doesn't pick that up.

lgruen (approved):
Awesome, thank you!

cpg_infra/abstraction/gcp.py
illusional: The list instantiation was to duplicate the list to avoid adding it to an existing list - I've refactored this to just be clearer and added a comment.

PR #16, opened 2022-11-02
Implement container registry for a dataset

illusional:
Will be useful for a universially dependent `cpg-common` dataset with an artifact registry, eventually removing the support for specifying cpg-common manually.

\+ Import the pre-existing fewgenomes repository.

### DIFF
--- a/cpg_infra/driver.py
+++ b/cpg_infra/driver.py
@@ -988,7 +988,6 @@ class CpgDatasetInfrastructure:
                     membership=ContainerRegistryMembership.APPEND,
                 )
 
-
     def setup_inherited_container_registries(self):
         """
         Setup permissions for cpg-common + analysis-runner artifact registries
--- a/cpg_infra/driver.py
+++ b/cpg_infra/driver.py
@@ -960,7 +960,41 @@ class CpgDatasetInfrastructure:
         + cpg-common container registries
         :return:
         """
+        self.setup_dataset_container_registry()
+        self.setup_inherited_container_registries()
 
+    def setup_dataset_container_registry(self):
+        """
+        If required, setup a container registry for a dataset
+        :return:
+        """
+        if not self.dataset_config.create_container_registry:
+            return
+
+        # mostly because this current format requires the project_id
+        custom_container_registry = self.infra.create_container_registry('images')
+        for kind, account in self.access_level_groups:
+            self.infra.add_member_to_container_registry(
+                f'test-images-reader-in-container-registry',
+                registry=custom_container_registry,
+                member=account,
+                membership=ContainerRegistryMembership.READER,
+            )
+            if kind in ('standard', 'full'):
+                self.infra.add_member_to_container_registry(
+                    f'{kind}-images-writer-in-container-registry',
+                    registry=custom_container_registry,
+                    member=account,
+                    membership=ContainerRegistryMembership.APPEND,
+                )
+
+
+    def setup_inherited_container_registries(self):
+        """
+        Setup permissions for cpg-common + analysis-runner artifact registries
+        """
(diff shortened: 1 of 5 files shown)


cpg_infra/driver.py
```diff
@@ -960,7 +960,40 @@ def setup_container_registry(self):
...
+                registry=custom_container_registry,
+                member=account,
+                membership=ContainerRegistryMembership.READER,
+            )
+            if kind in ('standard', 'full'):
+                self.infra.add_member_to_container_registry(
+                    f'{kind}-images-writer-in-container-registry',
+                    registry=custom_container_registry,
+                    member=account,
+                    membership=ContainerRegistryMembership.APPEND,
```
lgruen, line 988: Not directly related to this PR, but I just noted that `ContainerRegistryMembership.APPEND` is implemented as `role = 'roles/artifactregistry.writer'`, which seems to imply more than just appending? (I.e. also overwrite + delete?)
https://github.com/populationgenomics/cpg-infrastructure/blob/97d72179727f27a991a318ffae3e7b30990fe92c/cpg_infra/abstraction/gcp.py#L487-L488

This seems different from what we typically allow for `standard` (although we might want to revise that anyhow).

cpg_infra/driver.py
```diff
@@ -960,7 +960,40 @@ def setup_container_registry(self):
...
+                )
+
+    def setup_inherited_container_registries(self):
```
lgruen, line 991: If this is mostly meant to handle our existing setup temporarily, maybe "legacy" is the better description?

```suggestion
    def setup_legacy_container_registries(self):
```

If you want to change the implementation later to support access to repositories of dataset dependencies, maybe `dependent` could be a better description?

cpg_infra/driver.py
```diff
@@ -991,14 +1024,14 @@ def setup_container_registry(self):
...
-                    member=account,
-                    membership=ContainerRegistryMembership.APPEND,
-                )
+        for kind in ('full', 'standard'):
+            self.infra.add_member_to_container_registry(
+                f'{kind}-images-writer-in-cpg-common',
+                registry=self.config.gcp.common_artifact_registry_name,
+                project=self.config.gcp.common_artifact_registry_project,
+                member=self.access_level_groups[kind],
+                membership=ContainerRegistryMembership.APPEND,
```
lgruen, line 1033: Ditto.

stack/Pulumi.fewgenomes.yaml
```diff
@@ -5,6 +5,7 @@ config:
   datasets:gcp_hail_service_account_full: fewgenomes-full-915@hail-295901.iam.gserviceaccount.com
   datasets:gcp_hail_service_account_standard: fewgenomes-standard-737@hail-295901.iam.gserviceaccount.com
   datasets:gcp_hail_service_account_test: fewgenomes-test-126@hail-295901.iam.gserviceaccount.com
+  datasets:create_container_registry: "true"
```
lgruen, line 8: Or maybe remove this if you want to delete the registry anyhow? You might have to delete all images manually first.

lgruen (approved):

cpg_infra/driver.py
illusional: Agreed, renamed to `writer`, although I believe you can't overwrite (as no delete) with writer: https://cloud.google.com/artifact-registry/docs/access-control

cpg_infra/driver.py
illusional: Yep agreed, renamed to legacy.

stack/Pulumi.fewgenomes.yaml
illusional: Removed and deleted the artifact repos

PR #21, opened 2022-11-22
Add common stack 

illusional:
Also:

- rename 'reference' common set
- remove cpg-common artifact registry as common dependency

This script manages the artifact registry permissions (which was imported manually).

### DIFF
--- a/tokens/repository-map.json
+++ b/tokens/repository-map.json
@@ -17,6 +17,7 @@
     "circa": [
         "automated-interpretation-pipeline"
     ],
+    "common": [],
     "constraint": [
         "structural-constraint"
     ],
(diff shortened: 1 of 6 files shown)


tokens/repository-map.json
```diff
@@ -17,6 +17,7 @@
     "circa": [
         "automated-interpretation-pipeline"
     ],
+    "common": [],
```
vladsavelyev, line 20: Should we add `references` and `images` here?

vladsavelyev (approved):

tokens/repository-map.json
illusional: I have not created hail service accounts for the `common` dataset, so it's mostly just to ensure anything that's using a cached group members is actually picked up correctly - doesn't really do much else than that.

PR #24, opened 2022-11-23
PoC: AWS support

illusional:
For ABACBS presentation.

Noting that after #33, this more possible as we're able to completely resolve group members (as AWS does not support nesting groups) - though groups don't control access to resources, so maybe it's a moot point.


illusional:
Closing as very out of date, would be easier to restart this PR.

PR #33, opened 2023-01-02
Groups refactor + membership caching

illusional:
Plan:

- [x] Add group provider for managing groups + memberships, this allows for group memberships to be completely resolved in Pulumi
- [x] Change group cache to use buckets instead of secrets
- [x] Remove access-group-cache polling function
- [x] Create new groups per predefined spec
  - [x] User groups
  - [x] Machine groups (`test-read`, `test-write`, `main-list`, `main-read`, `main-write`)
  - [x] Introduce read-only depends structure
- [x] Add general group members to private repository
  - [x] Add mechanism for including those group members in this repo
  
This PR removes the OutputProvider, and replaces it with a better mechanism for accessing infrastructure between stacks - including more central infrastructure that can be globally accessed by any stack (in this case, the group-member-cache bucket). I think in future this will be a powerful way to manage infrastructure centrally through this repo.


illusional:
@lgruen, I think this will be tricky to deploy, without potentially removing all manually added members from groups. There are a lot of of group membership changes (to help deal with the read-only depends on).

Would love your thoughts :)

cpg_infra/driver.py
```diff
@@ -248,60 +479,150 @@ def setup_access_groups(self):
...
+
+        # transitive storage groups
+        self.test_read_group.add_member('test-full-in-test-read', self.test_full_group)
+        self.test_full_group.add_member(
+            'analysis-group-in-test-full', self.analysis_group
+        )
+        self.test_full_group.add_member('full-in-test-full', self.full_group)
+        self.test_full_group.add_member('test-in-test-full', self.test_group)
+
+        self.main_list_group.add_member(
```
illusional, line 551: Should standard be in test-full?

cpg_infra/driver.py
```diff
@@ -240,68 +461,196 @@ def deployment_accounts_by_access_level(self):
...
+
+        def compute_hash(s):
+            # add a salt so you can't track members through the resource names
+            # between different stacks - a small obfuscation.
+            return xxhash.xxh32((self.dataset_config.dataset + s).encode()).hexdigest()
+
+        for group in groups:
+            group_name = group.name.removeprefix(self.dataset_config.dataset + '-')
+            for member in d.get(group_name, []):
+                h = compute_hash(member)
```
lgruen, line 504: Do we want to use a more readable (but nevertheless sanitized) version here? Something like mapping `micfra@mail.com.au` --> `m****a@mail.com.au` or similar? The deployment diffs would become more useful that way.

stack/production.yaml
```diff
@@ -341,7 +341,7 @@ seqr:
     project: seqr-308602
 sgs-somatic-mtn:
   archive_age: 90
-  depends_on:
+  depends_on_readonly:
```
lgruen, line 344: Should we also change this for the `kccg` datasets?

stack/requirements.txt
```diff
@@ -5,3 +5,4 @@ requests==2.27.1
 PyYAML==5.4.1
 click==8.0.3
 cpg-utils
+xxhash==3.2.0
```
lgruen, line 8: Probably not needed if we change the obfuscation.

lgruen (approved):

stack/requirements.txt
illusional: Keeping to avoid clashes

cpg_infra/driver.py
illusional: I've switched this to use the initials, followed by the hash to avoid clashes and promote readability

cpg_infra/driver.py
```diff
@@ -173,6 +172,10 @@ def __init__(
         ] = defaultdict()
 
     def resolve_dataset_order(self):
+        """
+        This is now obsolete because we setup all datasets, and then deploy
+        so they all exist before the actual deploy.
+        """
```
lgruen, line 178: If it's obselete, can we remove it? :)

cpg_infra/driver.py
```diff
@@ -173,6 +172,10 @@ def __init__(
         ] = defaultdict()
 
     def resolve_dataset_order(self):
+        """
+        This is now obsolete because we setup all datasets, and then deploy
```
lgruen, line 176:
```suggestion
        This is now obsolete because we set up all datasets, and then deploy
```

cpg_infra/driver.py
```diff
@@ -1707,6 +1807,7 @@ def setup_legacy_container_registries(self):
         }
 
         for kind, account in kinds.items():
+            # TODO, this can be solved on high level by adding common-images-reader to analysis-runner container registry
```
lgruen, line 1810: Did you mean to resolve this?

lgruen (approved):

tokens/repository-map.json
```diff
@@ -28,6 +28,7 @@
     "constraint": [
         "structural-constraint"
     ],
+    "cpginfradev": ["cpg-infrastructure"],
```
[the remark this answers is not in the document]
illusional: Sort of, but better to leave out

cpg_infra/driver.py
illusional: I wasn't going to, but have since resolved.

cpg_infra/driver.py
```diff
@@ -249,13 +254,22 @@ def deploy_datasets(self):
                 dataset_infra.main()
 
     def finalize_groups(self):
+        def _process_members(members):
+            # use .sort twice because python sort is in place and stable
```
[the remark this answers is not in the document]
illusional: Yeah better idea, actioned.

cpg_infra/driver.py
illusional: I think it influences the pulumi create order, which is useful when spinning up things - but it doesn't really matter. I've left it because it's a useful code snippet to have around.

PR #35, opened 2023-01-16
Update repository-map.json

MattWellie:
Opening a request to have AIP access the Broad-RGP data as it sits in the current bucket (gs://cpg-broad-rgp-main (-upload))

Is this a viable solution? Are the permissions configured appropriately for billing etc. so that we'll be able to get going on this data ASAP?


illusional (approved):
:o 
I need to learn more about the processors, thanks for fixing this!

PR #40, opened 2023-01-31
Setup billing management system in Azure in cpg-infra

violetbrina:
Create billing management functions in Azure

Sub-tasks:
- [ ] Create subscription level deployment of a function "stop-compute" that applies a policy to any child resource group able to prevent new compute (using a policy) and pause/stop existing vms in the resource group
- [ ] Add dataset level budget triggers (either a pub/sub like equivalent or budget trigger) to call the global "stop-compute" function when budget is exceeded
- [ ] Add a follow up trigger/dataset function that when a budget is increased/updated (using pulumi) the "stop-compute" function is called to re-enable compute on that resource group (change the policy back to allow vms again)
- [ ] Test the setup to ensure that when budget is exceeded vms are prevented/stopped
- [ ] Test the setup to ensure that when the budget is increased through pulumi vms are allowed


illusional:
Hey @violetbrina, going to dismiss my review until you're ready for me again

PR #48, opened 2023-02-06
Ensure trailing newline for members cache

illusional:
Untested, but something like this should work.

### DIFF
--- a/cpg_infra/driver.py
+++ b/cpg_infra/driver.py
@@ -265,16 +265,16 @@ class CPGInfrastructure:
                 if group.cache_members and isinstance(infra, GcpInfrastructure):
                     _members = self.group_provider.resolve_group_members(group)
                     member_ids = [infra.member_id(m) for m in _members]
-                    members_contents = ''
+                    members_contents = '\n'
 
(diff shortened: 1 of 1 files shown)


cpg_infra/driver.py
```diff
@@ -265,16 +265,16 @@ def _process_members(members):
...
                     if len(member_ids) > 0:
                         if all(isinstance(m, str) for m in member_ids):
-                            members_contents = _process_members(member_ids)
+                            members_contents = _process_members(member_ids) + '\n'
                         else:
                             members_contents = (
                                 pulumi.Output.all(*member_ids)
                                 .apply(_process_members)
-                                .apply(lambda value: value or '')
+                                .apply(lambda value: (value or '') + '\n')
```
lgruen, line 277: Probably don't even need the newlines here, as we just need to make the "empty" case not fail.

But maybe you want it for consistency, which is of course fine too :)

lgruen, line 277: `(value or '\n')` might also work, but either way is fine.

lgruen (approved):
Awesome, thanks!

PR #49, opened 2023-02-07
Add autoclass config option

lgruen:


cpg_infra/abstraction/azure.py
```diff
@@ -270,8 +270,12 @@ def create_bucket(
         unique: bool = False,
         requester_pays: bool = False,
         versioning: bool = True,
+        autoclass: bool = False,
         project: str = None,
     ) -> Any:
+        # Autoclass is not supported on Azure yet.
```
illusional, line 276: This will crash immediately, because there's no difference between the gcp and azure calls. I think it would be reasonable to ignore, or throw a warning, but assert will block the deploy.

cpg_infra/driver.py
```diff
@@ -1187,26 +1191,33 @@ def main_tmp_bucket(self):
...
 
     @cached_property
     def main_upload_buckets(self) -> dict[str, Any]:
         main_upload_undelete = self.infra.bucket_rule_undelete(days=30)
         main_upload_buckets = {
             'main-upload': self.infra.create_bucket(
-                'main-upload', lifecycle_rules=[main_upload_undelete]
+                'main-upload',
+                lifecycle_rules=[main_upload_undelete],
+                autoclass=self.dataset_config.autoclass,
```
illusional, line 1220: Do we want things in upload to move to archive? I think things in main-upload should start throwing warnings if they haven't been actioned in a reasonable amount of time.

illusional (changes requested):

cpg_infra/abstraction/azure.py
lgruen: Oh, interesting. So we always create the buckets for all datasets on both GCP and Azure? I thought we'd want some sort of selector at the dataset level.

cpg_infra/driver.py
lgruen: Well, it doesn't hurt to have Autoclass enabled. There certainly have been files in upload for some period of time, but I agree that ideally we'd also have an alert. That's a bit orthogonal though.

cpg_infra/abstraction/azure.py
illusional: There is a selector for the dataset level, but cpg-common is already spun up across both.

illusional (approved):
In the future, do you want the default to be true?

cpg_infra/abstraction/azure.py
lgruen: Of course, good point.

lgruen: Changed this to a warning now.

lgruen:
> In the future, do you want the default to be true?

I think we should just get rid of the setting completely once we've finished the migration.

PR #58, opened 2023-02-08
Add group prefix + configure for dev

illusional:


cpg_infra/config.py
```diff
@@ -166,6 +166,7 @@ class GCP(DeserializableDataclass):
     # delete, which can end up with missing permissions. To implement the first phase
     # (delete), simply change this to 'True', then revert to reapply group memberships
     disable_group_memberships: bool = False
+    group_prefix: str = ''
```
lgruen, line 169: Could you add a comment here describing when it's useful to set this?
(Could maybe default this to `None` too?)

lgruen (approved):
Very helpful, thanks!

stack/dev.yaml
```diff
@@ -1,4 +1,5 @@
 cpginfradev:
+  autoclass: true
```
lgruen, line 2: Accidentally included in this PR?

illusional: Addressing both issues affecting dev stack ;)

cpg_infra/config.py
illusional: Resolved :)

PR #64, opened 2023-02-12
Add data manager to credentials generator

illusional:
Context: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1676239341416309?thread_ts=1675743196.294449&cid=C030X7WGFCL

### DIFF
--- a/cpg_infra/abstraction/base.py
+++ b/cpg_infra/abstraction/base.py
@@ -58,7 +58,7 @@ class MachineAccountRole(Enum):
     """Roles for users to interact with machine-account"""
 
     ACCESS = 'access'
-    CREDENTIALS_GENERATOR = 'credentials-generator'
+    ADMIN = 'admin'
 
 
 class CloudInfraBase(ABC):
--- a/cpg_infra/abstraction/base.py
+++ b/cpg_infra/abstraction/base.py
@@ -54,6 +54,13 @@ class ContainerRegistryMembership(Enum):
     WRITER = 'writer'
 
 
+class MachineAccountRole(Enum):
+    """Roles for users to interact with machine-account"""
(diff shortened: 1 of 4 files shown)


cpg_infra/driver.py
```diff
@@ -727,6 +728,16 @@ def setup_web_access_group_memberships(self):
             member=self.analysis_group,
         )
 
+    def setup_service_account_generation_permission(self):
+        if not isinstance(self.infra, GcpInfrastructure):
+            return
+
+        self.infra.add_project_role(
```
lgruen, line 735: Can we add this on the service account itself instead of the whole project?

lgruen (changes requested):

cpg_infra/driver.py
illusional: Yep, definitely makes sense

cpg_infra/abstraction/base.py
```diff
@@ -333,8 +345,13 @@ def create_machine_account(
...
-        self, resource_key: str, machine_account, member, project: str = None
+    def add_member_to_machine_account_role(
+        self,
+        resource_key: str,
+        machine_account,
+        member,
+        role: MachineAccountRole,
+        project: str = None,
     ) -> Any:
         print(f'Allow {member} to access {machine_account}')
```
lgruen, line 356: Did you mean to just have a simple `pass` here?

lgruen (approved):

cpg_infra/abstraction/base.py
illusional: This is a DEV CloudInfraBase, so it's supposed to log out what it's doing.

PR #65, opened 2023-02-13
Migrate stack files to cpg-infra-private

illusional:
https://github.com/populationgenomics/cpg-infrastructure-private/pull/8


requirements.txt
```diff
@@ -3,4 +3,5 @@ pulumi-gcp==6.49.0
 pulumi-azure-native==1.80.0
 pulumi-azuread==5.28.0
 toml==0.10.2
+xxhash==3.2.0
```
lgruen, line 6: Was this intentional?

cpg_infra/abstraction/azure.py
```diff
@@ -48,9 +48,9 @@ def __init__(
         self.storage_account_lifecycle_rules = []
         self.storage_account_undelete_rule = None
 
-        data = az.authorization.get_client_config()
-        self.subscription_id = '/subscriptions/' + data.subscription_id
-        self.tenant = data.tenant_id
+        # data = az.authorization.get_client_config()
```
lgruen, line 51: This seems unrelated?

lgruen (changes requested):

requirements.txt
illusional: I included this and the below change as they made migrating the state easier.

This, so I can install directly with pip as it was missing. And the below change because it helped for Azure debugging, but then realised it wasn't all relevant anyway.

I can move these to a separate PR, but they do relate to the stack migration to infra private

cpg_infra/abstraction/azure.py
illusional: This was initially failing with some related az auth issue, but then it occurred through this migration process that it's not needed. I've removed the line instead of commenting it out.

lgruen (approved):

PR #75, opened 2023-02-16
Remove grace period for archive storage

lgruen:
See companion PR https://github.com/populationgenomics/cpg-infrastructure-private/pull/21.

### DIFF
--- a/cpg_infra/driver.py
+++ b/cpg_infra/driver.py
@@ -1075,10 +1075,7 @@ class CPGDatasetInfrastructure:
     def archive_bucket(self):
         return self.infra.create_bucket(
             'archive',
-            lifecycle_rules=[
-                self.infra.bucket_rule_archive(days=self.dataset_config.archive_age),
-                self.infra.bucket_rule_undelete(),
-            ],
+            lifecycle_rules=[self.infra.bucket_rule_undelete()],
             autoclass=False,  # Manually managed cold tier.
         )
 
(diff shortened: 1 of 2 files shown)


cpg_infra/driver.py
```diff
@@ -1075,10 +1075,7 @@ def setup_storage_archive_bucket_permissions(self):
     def archive_bucket(self):
         return self.infra.create_bucket(
             'archive',
-            lifecycle_rules=[
-                self.infra.bucket_rule_archive(days=self.dataset_config.archive_age),
-                self.infra.bucket_rule_undelete(),
-            ],
+            lifecycle_rules=[self.infra.bucket_rule_undelete()],
```
illusional, line 1078: How does this actually be in the cold tier if autoclass is false and tehre are no lifecycle rules or bucket tier options?

illusional (changes requested):

cpg_infra/driver.py
lgruen: Of course you're completely right, this PR was pretty nonsensical. It's easy to change the `storage_class` of a bucket in GCS (https://www.pulumi.com/registry/packages/gcp/api-docs/storage/bucket/#inputs), but unfortunately that concept doesn't seem to exist in Azure, as far as I can tell (https://learn.microsoft.com/en-us/azure/storage/blobs/access-tiers-overview#changing-a-blobs-access-tier).

Hence I'll close this PR and instead submit one where we default to an archive age / grace period of 0 days.

PR #76, opened 2023-02-16
Allow interview exercises to access fewgenomes

lgruen:


lgruen:
Nevermind, this doesn't work, as it's not a public repo.

PR #80, opened 2023-02-21
Default-enable Autoclass

lgruen:
**Note**: depends on https://github.com/populationgenomics/cpg-infrastructure-private/pull/34 being merged. Do not deploy beforehand.


illusional (approved):
Is this defaulted in the new_stack script already?

lgruen:
`new_stack` doesn't set it at all, which is why I'd like to change the default. As a next step, I'll remove all the `autoclass: true` lines from `production.yaml`.

PR #87, opened 2023-02-28
Sort toml files to stabilise outputs

illusional:


cpg_infra/billing_aggregator/driver.py
```diff
@@ -187,6 +187,7 @@ def create_cloud_function(
         event_type='google.cloud.pubsub.topic.v1.messagePublished',
         trigger_region='australia-southeast1',
         pubsub_topic=pubsub_topic.id,
+        retry_policy='RETRY_POLICY_DO_NOT_RETRY',
```
illusional, line 190: Unrelated, I know - but it was convenient to put it here to stop updates after a refresh

lgruen (approved):

PR #95, opened 2023-03-07
Explicitly include .txt / .json files in MANIFEST

illusional:
Not including this means the requirements.txt / json files required for the billing functions are not included when installed in the deploy action in infra-private.


lgruen (approved):
🏅

PR #97, opened 2023-03-07
Simplify ID on hail-derived billing entries to avoid resource changes in the future

illusional:
As discussed in our catch-up :)

### DIFF
--- a/cpg_infra/billing_aggregator/aggregate/hail.py
+++ b/cpg_infra/billing_aggregator/aggregate/hail.py
@@ -77,61 +77,66 @@ def get_finalised_entries_for_batch(batch: dict) -> List[Dict]:
     batch_id = batch['id']
     dataset = batch['billing_project']
     currency_conversion_rate = utils.get_currency_conversion_rate_for_time(start_time)
+    attributes = batch.get('attributes', {})
+    batch_url = utils.HAIL_UI_URL.replace('{batch_id}', str(batch_id))
 
-    resource_usage = defaultdict(float)
-    resource_cost = defaultdict(float)
     for job in batch['jobs']:
-        for batch_resource, usage in job['resources'].items():
-            resource_usage[batch_resource] += usage
+        for batch_resource, raw_cost in job['cost'].items():
+            if batch_resource.startswith('service-fee'):
+                continue
 
-        for batch_resource, cost in job['cost'].items():
-            resource_cost[batch_resource] += cost
+            job_id = job['job_id']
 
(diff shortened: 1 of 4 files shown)


cpg_infra/billing_aggregator/aggregate/hail.py
```diff
@@ -112,13 +112,14 @@ def get_finalised_entries_for_batch(batch: dict) -> List[Dict]:
 
         cost = utils.get_total_hail_cost(currency_conversion_rate, raw_cost=raw_cost)
         usage = resource_usage.get(batch_resource, 0)
+        # 2023-03-07 mfranklin: I know this key isn't unique, but to avoid issues
+        # with changing the resource_id again, we'll only use the batch_id as the key
+        # as it's sensible for us to assume that all the entries exist if one of the
+        # entries exists
```
lgruen, line 118: Is that really true though if some jobs are running much longer than others? Should the key at least also include the job ID?

cpg_infra/billing_aggregator/aggregate/seqr.py
```diff
@@ -161,8 +161,13 @@ def get_finalised_entries_for_batch(
                 currency_conversion_rate, raw_cost=raw_cost
             )
             raw_usage = job['resources'].get(batch_resource, 0)
+
+            # Distribute the remaining cost across all datasets proportionately
```
lgruen, line 165:
```suggestion
            # Distribute the remaining cost across all datasets proportionally
```

cpg_infra/billing_aggregator/aggregate/seqr.py
```diff
@@ -161,8 +161,13 @@ def get_finalised_entries_for_batch(
...
             for dataset, (fraction, dataset_size) in prop_map.items():
-                # Distribute the remaining cost across all datasets proportionately
+                # 2023-03-07 mfranklin: I know this key isn't unique, but to avoid
+                # issues with changing the resource_id again, we'll only use the
+                # batch_id as the key as it's sensible for us to assume that all the
```
lgruen, line 169: Here and below, the comment says "only use the batch_id", but the implementation actually includes the `job_id` too.

lgruen (approved):

cpg_infra/billing_aggregator/aggregate/hail.py
illusional: This one aggregates the whole batch (instead of per job), because the batch is completed + closed. 
We only distribute seqr (right now).

PR #104, opened 2023-03-20
Bump memory for GCP aggregator

illusional:
Memory issues, bigger to be safer

### DIFF
--- a/cpg_infra/billing_aggregator/driver.py
+++ b/cpg_infra/billing_aggregator/driver.py
@@ -242,11 +242,11 @@ class BillingAggregator:
         )
 
         for function in self.config.billing.aggregator.functions:
-            memory = '256M'
-            cpu = None
+            # Balance CPU by this table:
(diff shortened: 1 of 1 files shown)


violetbrina (approved):

cpg_infra/billing_aggregator/driver.py
```diff
@@ -242,11 +242,11 @@ def setup_aggregator_functions(self):
         )
 
         for function in self.config.billing.aggregator.functions:
-            memory = '256M'
-            cpu = None
+            # Balance CPU by this table:
+            # https://cloud.google.com/functions/docs/configuring/memory
+            memory = '1024M'
+            cpu = 0.583
```
lgruen, line 248: Was that odd number intentional?

lgruen (approved):

cpg_infra/billing_aggregator/driver.py
illusional: Yep, lines up with the CPU in the table

PR #105, opened 2023-03-20
Add storage visualization automation

lgruen:
Example: https://test-web.populationgenomics.org.au/common/storage_visualization/23-03-21/treemap.html

### DIFF
--- a/storage_visualization/submit.py
+++ b/storage_visualization/submit.py
@@ -34,7 +34,7 @@ def main():
             f'storage_visualization_{datetime.date.today().strftime("%y-%m-%d")}',
             '--description',
             'Storage visualization',
-            'main.py',
+            'storage_visualization/main.py',
         ]
         + datasets
     )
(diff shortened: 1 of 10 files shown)


storage_visualization/submit.py
```diff
@@ -0,0 +1,46 @@
...
+        print('Usage: submit.py <config.yaml>')
+        sys.exit(1)
+
+    with open(sys.argv[1], encoding='utf-8') as f:
+        config = yaml.safe_load(f)
+
+    # Generate a list of all datasets to invoke main.py with.
+    datasets = sorted(list(set(config.keys()) - {ALL_DATASETS}))
+
+    subprocess.check_output(
```
illusional, line 26: Is the analysis-runner dependency installed? If so, would be cleaner to call the CLI method directly, https://github.com/populationgenomics/analysis-runner/blob/c4f8f91c6fb0664923d29d11ea356865df0001ea/analysis_runner/cli_analysisrunner.py#L98

illusional (approved):

PR #107, opened 2023-03-23
Add storage_visualization.yaml reqs

illusional:
- ~~Running once as a test~~ Test failed as expected

### DIFF
--- a/.github/workflows/storage_visualization.yaml
+++ b/.github/workflows/storage_visualization.yaml
@@ -6,6 +6,7 @@ on:
     - cron: "0 0 * * 0"
   # Run on demand.
   workflow_dispatch:
+  push: ['add-storage-viz-reqs']
 
 jobs:
   submit_batch:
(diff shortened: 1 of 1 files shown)


.github/workflows/storage_visualization.yaml
```diff
@@ -29,5 +31,12 @@ jobs:
           credentials_json: "${{ secrets.GCP_STORAGE_VISUALIZATION_CREDENTIALS }}"
           project_id: cpg-common
 
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+          cache: "pip" # caching pip dependencies
+
+      - run: pip install -r storage_visualization/requirements.txt analysis-runner
```
lgruen, line 39: `submit.py` itself doesn't actually need the other dependencies. But might not hurt either way...

```suggestion
      - run: pip install analysis-runner
```

.github/workflows/storage_visualization.yaml
```diff
@@ -6,6 +6,8 @@ on:
     - cron: "0 0 * * 0"
   # Run on demand.
   workflow_dispatch:
+  push:
+    branches: ['add-storage-viz-reqs']
```
lgruen, line 10: Debugging remnant?

```suggestion
```

lgruen (approved):

.github/workflows/storage_visualization.yaml
illusional: Just removed, didn't mean to tag you before I had removed that.

PR #121, opened 2023-04-17
Post preview image for storage visualization

lgruen:
Depends on https://github.com/populationgenomics/cpg-utils/pull/138.

### DIFF
--- a/storage_visualization/main.py
+++ b/storage_visualization/main.py
@@ -6,6 +6,7 @@ import sys
 
 # See requirements.txt for why we're disabling the linter warnings here.
 import hailtop.batch as hb  # pylint: disable=import-error
+from cloudpathlib import AnyPath
 
 from cpg_utils.config import get_config
 from cpg_utils.git import (
@@ -19,7 +20,7 @@ from cpg_utils.hail_batch import (
     remote_tmpdir,
     output_path,
 )
-from cpg_utils.slack import send_message
+from cpg_utils.slack import upload_file
 
 
 def prepare_job(job, clone_repo):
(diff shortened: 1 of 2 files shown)


storage_visualization/main.py
```diff
@@ -19,12 +20,16 @@
...
+
+DOCKER_IMAGE = (
+    'australia-southeast1-docker.pkg.dev/cpg-common/images/storage-visualization:latest'
+)
 
 
 def prepare_job(job, clone_repo):
     """Sets up the given job to run scripts in the same repository."""
-    job.image(get_config()['workflow']['driver_image'])
+    job.image(DOCKER_IMAGE)
```
illusional, line 32: Can you only use the `kaleido` image just for the job that needs it - that way we don't have to consistently keep it up to date.

.github/workflows/storage_visualization.yaml
```diff
@@ -29,6 +29,15 @@ jobs:
           credentials_json: "${{ secrets.GCP_STORAGE_VISUALIZATION_CREDENTIALS }}"
           project_id: cpg-common
 
+      - name: "gcloud docker auth"
```
illusional, line 32: I've commented below, but I think it would be better to only use the kaleido image on the job where it's needed, then we don't need to push a docker image each time it runs.

illusional (changes requested):

.github/workflows/storage_visualization.yaml
lgruen: Are you suggesting to build the Kaleido image manually only? Then it'll go out of date with the driver image. This runs only once a week and the additional layer shouldn't cost much -- we also override the previous tag, so this won't accumulate versions.

storage_visualization/main.py
lgruen: See below -- isn't it better to make sure everything is consistently up-to-date?

.github/workflows/storage_visualization.yaml
illusional: If it's only kaleido, why not freeze it at that version? I'll accept it either way (because IMO it doesn't matter too much either way), just mostly wanted to prompt if it was necessary.

lgruen: Oh, simply because we're not just using kaleido in isolation, but these jobs also write to blob store, upload to Slack, etc. I agree that it likely doesn't make much of a difference in practice, but just seemed more robust to always update, in case our utility libraries change.

illusional (approved):

PR #128, opened 2023-05-05
Membership approval refresh (API changes)

illusional:
Paired with: https://github.com/populationgenomics/cpg-infrastructure-private/pull/117

This PR introduces changes to support a `user.yaml` map that includes extra attributes like an azure ID, and hail batch username, some of which were hacked on as part of the config previously.

This _much_ more cleanly represents what attributes are required for including users in our infrastructure.


illusional:
@daniaki, this PR has preview errors, because it requires the cpg-infrastructure-private changes to be deployed. I've confirmed these locally, and will deploy this locally to manage the deploy.

cpg_infra/driver.py
```diff
@@ -661,38 +657,33 @@ def setup_externally_specified_members(self):
...
+                member = self.config.users.get(member_id)
+                if not member:
+                    raise ValueError(f'Member {member_id} not found in config')
+
+                h = self.compute_hash(
+                    dataset=self.dataset_config.dataset,
+                    member=member.id,
+                    cloud=self.infra.name(),
+                )
+                if cloud_user := member.clouds[self.infra.name()]:
```
daniaki, line 670: Could this raise a `KeyError` if the infra name is not in a member's clouds dict, or is `member.clouds` is initialised to contain all expected infra names?

daniaki (approved):

cpg_infra/driver.py
illusional: We only want to add someone to a cloud where we know the "id" we can use to reference them. In Google Cloud this is easy, it's just their email (encoded as gcp.id in the user.yaml). For azure, this is a specific object ID they get when they're guest-ed into our tenant. So we only want to add them if we know that ID. 

It could lead to some awkward debugging, "why isn't this person in the azure group", but I don't want to add logging because it's the more common case that they're not added to an equivalent azure resource. 

Same vibes for Hail below, only add them if we know they're hail ID for the specific cloud.

PR #131, opened 2023-05-12
Split dataset storage prefix by cloud

illusional:
When I tore down our infrastructure, it didn't delete the accounts properly, so now `cpgcommon` is permanently unavailable, and only allowed `[a-z][a-z0-9]{3,24}` so we're quite stuck except to append number or pick a new prefix


violetbrina:
Ah I see. That's unfortunate. I'm assuming there's no way to tear down the storage account now? Oh well. Probably not a bad thing that the infra code is slightly more robust and can work in more cases anyway.

violetbrina (approved):

PR #132, opened 2023-05-19
Allow external members in Google Groups

lgruen:
Currently enabled for _all_ groups, but we could easily make that more customizable.

### DIFF
--- a/cpg_infra/abstraction/gcp.py
+++ b/cpg_infra/abstraction/gcp.py
@@ -21,7 +21,7 @@ from cpg_infra.abstraction.base import (
     BUCKET_DELETE_INCOMPLETE_UPLOAD_PERIOD_IN_DAYS,
     MachineAccountRole,
 )
-from cpg_infra.abstraction.group_settings import GroupSettings
+from cpg_infra.abstraction.google_group_settings import GoogleGroupSettings
 from cpg_infra.config import CPGDatasetConfig, CPGInfrastructureConfig
 
 
@@ -503,7 +503,7 @@ class GcpInfrastructure(CloudInfraBase):
         )
         # Allow domain-external members in the group.
         GoogleGroupSettings(
-            f'{name}-group-settings',
+            self.get_pulumi_name(name + '-group-settings'),
             group_email=mail,
             settings={'allowExternalMembers': 'true'},
             opts=pulumi.resource.ResourceOptions(depends_on=[group]),
--- a/cpg_infra/abstraction/gcp.py
+++ b/cpg_infra/abstraction/gcp.py
@@ -21,6 +21,7 @@ from cpg_infra.abstraction.base import (
     BUCKET_DELETE_INCOMPLETE_UPLOAD_PERIOD_IN_DAYS,
     MachineAccountRole,
 )
+from cpg_infra.abstraction.group_settings import GroupSettings
 from cpg_infra.config import CPGDatasetConfig, CPGInfrastructureConfig
 
 
@@ -492,7 +493,7 @@ class GcpInfrastructure(CloudInfraBase):
 
     def create_group(self, name: str) -> Any:
         mail = f'{name}@{self.config.gcp.groups_domain}'
-        return gcp.cloudidentity.Group(
+        group = gcp.cloudidentity.Group(
             self.get_pulumi_name(name + '-group'),
             display_name=name,
             group_key=gcp.cloudidentity.GroupGroupKeyArgs(id=mail),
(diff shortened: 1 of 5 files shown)


cpg_infra/abstraction/gcp.py
```diff
@@ -492,14 +493,22 @@ def get_credentials_for_machine_account(self, resource_key, account):
...
             self.get_pulumi_name(name + '-group'),
             display_name=name,
             group_key=gcp.cloudidentity.GroupGroupKeyArgs(id=mail),
             labels={'cloudidentity.googleapis.com/groups.discussion_forum': ''},
             parent=f'customers/{self.config.gcp.customer_id}',
             opts=pulumi.resource.ResourceOptions(depends_on=[self._svc_cloudidentity]),
         )
+        # Allow domain-external members in the group.
+        GroupSettings(
+            f'{name}-group-settings',
```
illusional, line 506: Could be worth calling the `get-pulumi-name` to ensure the ID is definitely unique, but adding `gcp-` to the start should also be fine.

```suggestion
            f'gcp-{name}-group-settings',
```

cpg_infra/abstraction/group_settings.py
```diff
@@ -0,0 +1,113 @@
...
+
+from functools import cache
+import pulumi
+import pulumi.dynamic
+import google.auth
+import googleapiclient.discovery
+from google.auth.transport.requests import Request
+
+
+class GroupSettings(pulumi.dynamic.Resource):
```
illusional, line 14:
```suggestion
class GoogleGroupSettings(pulumi.dynamic.Resource):
```

cpg_infra/abstraction/group_settings.py
```diff
@@ -0,0 +1,113 @@
...
+    """A Pulumi dynamic resource for Google Groups settings."""
+
+    group_email: pulumi.Output[str]
+    settings: pulumi.Output[dict]
```
illusional, line 18: Is it worth defining the group settings as constants? Or could you direct link to the docs where those settings are defined?

cpg_infra/abstraction/group_settings.py
```diff
@@ -0,0 +1,113 @@
...
+        name: str,
+        group_email: str,
+        settings: dict,
+        opts: pulumi.ResourceOptions | None = None,
+    ):
+        super().__init__(
+            GroupSettingsProvider(),
+            name,
+            {
+                'group_email': pulumi.Output.from_input(group_email),
```
illusional, line 31: This is surprising Pulumi can't figure it out automatically

cpg_infra/abstraction/group_settings.py
```diff
@@ -0,0 +1,113 @@
...
+                'settings': pulumi.Output.from_input(settings),
+            },
+            opts,
+        )
+
+
+class GroupSettingsProvider(pulumi.dynamic.ResourceProvider):
```
illusional, line 38:
```suggestion
class GoogleGroupSettingsProvider(pulumi.dynamic.ResourceProvider):
```

cpg_infra/abstraction/group_settings.py
```diff
@@ -0,0 +1,113 @@
...
+            id_=f'google_group_settings::{group_email}', outs=outputs
+        )
+
+    def read(self, id_, props):
+        """Reads a Google Groups settings resource."""
+        group_email = props['group_email']
+        settings = props['settings']
+        current_settings = get_group_settings(group_email)
+        # The response contains *all* settings, so subset to relevant keys.
+        current_settings = {k: current_settings[k] for k in settings}
```
illusional, line 62: Should this be a `.get` in case a key suddenly ends up missing, which would then cause an update rather than an error?

cpg_infra/abstraction/gcp.py
```diff
@@ -492,14 +493,22 @@ def get_credentials_for_machine_account(self, resource_key, account):
...
+        group = gcp.cloudidentity.Group(
             self.get_pulumi_name(name + '-group'),
             display_name=name,
             group_key=gcp.cloudidentity.GroupGroupKeyArgs(id=mail),
             labels={'cloudidentity.googleapis.com/groups.discussion_forum': ''},
             parent=f'customers/{self.config.gcp.customer_id}',
             opts=pulumi.resource.ResourceOptions(depends_on=[self._svc_cloudidentity]),
         )
+        # Allow domain-external members in the group.
+        GroupSettings(
```
illusional, line 505:
```suggestion
        GoogleGroupSettings(
```

illusional (approved):
Very excited for this! A few nitpicks but otherwise looks great

cpg_infra/abstraction/group_settings.py
lgruen: Good idea to add a link. Done!

cpg_infra/abstraction/group_settings.py
lgruen: Thanks, it actually can -- it's a bit more concise now.

PR #136, opened 2023-06-05
Improve config parsing with respect to type annotations

illusional:


cpg_infra/config/deserializabledataclass.py
```diff
@@ -30,94 +30,110 @@ def __post_init__(self):
...
-    if ftype is None:
-        return None
-
-    if ftype in (list, dict) or get_origin(ftype) in (list, dict):
-        ftype_type = ftype if ftype in (list, dict) else get_origin(ftype)
-        value = config.get_object(fieldname)
+def try_parse_value_as_type(value, dtype):
+    """
+    Try to parse a value as a specific type, if it fails
+    return None
```
daniaki, line 71: Doc string might be misleading since an error will be raised for values that cannot be validated.

cpg_infra/config/deserializabledataclass.py
```diff
@@ -30,94 +30,110 @@ def __post_init__(self):
...
+        raise ValueError(message)
+
+    if isinstance(dtype, tuple) and len(dtype) == 1:
+        return try_parse_value_as_type(value, dtype[0])
+
+    if dtype is list or get_origin(dtype) is list:
+        if not isinstance(value, list):
+            raise ValueError(f'Expected list, got {type(value)} for {value!r}')
+        list_types = get_args(dtype)
+        if len(list_types) != 1:
```
daniaki, line 132: Is there a use-case where you would want to validate something like `list[a | b | c | ... ]` to make sure that list items only have types `a | b | c | ...` ?

cpg_infra/config/deserializabledataclass.py
```diff
@@ -30,94 +30,110 @@ def __post_init__(self):
...
+        dict_types = get_args(dtype)
+        if len(dict_types) != 2:
+            # no need for casting
+            return value
+        return {k: try_parse_value_as_type(v, dict_types[1]) for k, v in value.items()}
+    if dtype is set or get_origin(dtype) is set:
+        if not isinstance(value, (set, list)):
+            raise ValueError(f'Expected set, got {type(value)} for {value!r}')
+        set_types = get_args(dtype)
+        if len(set_types) != 1:
```
daniaki, line 151: Same comment as above on line 97: `if len(list_types) != 1`

daniaki (approved):
Left some small minor comments. I recently discovered a package called [pydantic](https://docs.pydantic.dev/latest/) which performs data validation on dataclasses by using type hint information. Seems pretty similar to what you're trying to achieve with `DeserializableDataclass`.

cpg_infra/config/deserializabledataclass.py
illusional: Yes! I've added a test to avoid regressions of this case :)

illusional: Ah I see the confusion, the type here from get_args for a list is always one element, and would come in as a UnionType which is caught below.

Some other types, like dict[str, str] have get_args returning 2 elements.

cpg_infra/config/deserializabledataclass.py
illusional: Nice catch, clarified

PR #139, opened 2023-06-07
Fix get start end from request

violetbrina:


cpg_infra/billing_aggregator/aggregate/utils.py
```diff
@@ -829,11 +829,11 @@ def get_start_and_end_from_request(
         logger.warning(f'Attributes could not be found in request: {request_data}')
         return None, None
 
-    if 'start' in request_data.get('attributes') or 'end' in request_data.get(
-        'attributes'
+    if 'start' in request_data.get('attributes', {}) or 'end' in request_data.get(
+        'attributes', {}
     ):
         request_data = request_data['attributes']
```
illusional, line 835: Maybe the walrus operator could be useful here:

```suggestion
    if attributes := request_data.get('attributes'):
        if 'start' in attributes or 'end' in attributes:
            request_data = attributes
```

illusional (approved):

PR #143, opened 2023-06-13
Metamist project creation + small dataset manager refactor

illusional:
Create the metamist project through Pulumi through another dynamic resource provider.
We only want to create _one_ per dataset, so now is a natural point to make the model support this.

Now, insert a new CPGDatasetInfrastructure in between CPGInfrastructure and CPGDatasetCloudInfrastructure. And make the generic Dataset one do the instantiation of the clouds. Simplified some logic at the top level, allows for single dataset infrastructure in one of the specific cloud providers (or cloud agnostic).

Introduces some simplifications, like accessing the "common" dataset to create infrastructure in, at the top level.

Adds `metamist` as a dependency (the new pypi version, even though it's not technically current, the method names for project won't change).

### DIFF
--- a/cpg_infra/abstraction/metamist.py
+++ b/cpg_infra/abstraction/metamist.py
@@ -32,26 +32,21 @@ class MetamistProjectProvider(pulumi.dynamic.ResourceProvider):
         name = props['project_name']
         create_test_project = props['create_test_project']
 
-        existing_project = get_project_by_name(name)
-        if existing_project:
-            return pulumi.dynamic.CreateResult(
-                id_=f'metamist-project::{name}::{existing_project["id"]}',
-                outs={
-                    'id': existing_project['id'],
-                    'name': name,
-                },
+        project = get_project_by_name(name)
+        if not project:
+            project = ProjectApi().create_project(
+                name=name,
+                dataset=name,
+                create_test_project=create_test_project,
             )
+        if not project:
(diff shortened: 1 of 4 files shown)


cpg_infra/abstraction/metamist.py
```diff
@@ -0,0 +1,99 @@
...
+class MetamistProjectProvider(pulumi.dynamic.ResourceProvider):
+    """Pulumi provider for creating a metamist project"""
+
+    def create(self, props) -> pulumi.dynamic.CreateResult:
+        name = props['project_name']
+        create_test_project = props['create_test_project']
+
+        existing_project = get_project_by_name(name)
+        if existing_project:
+            return pulumi.dynamic.CreateResult(
```
daniaki, line 37: Super super minor but I have nothing else to comment on so 🤷  - could compact this down to a single return statement by doing something like:

```python
# Check for existing project
project = get_project_by_name(name)
if not project:
    project = ProjectApi().create_project(...)
  
return pulumi.dynamic.CreateResult(...)
```

That way if the format of `id_` changes, it only needs to be updated in one spot.

daniaki (approved):

cpg_infra/abstraction/metamist.py
illusional: Yeah great idea, actioned!

PR #146, opened 2023-07-25
Fix bucket_rule_temporary() docstring

jmarshall:
This had been left as copy/pasted from the adjacent function definition.

This is indeed _n_ days after creation (one might guess it could instead be _n_ days after the last access):

* The _azure.py_ implementation uses `days_after_modification_greater_than`, which is unambiguous;
* The _gcp.py_ implementation sets `gcp.storage.BucketLifecycleRuleConditionArgs(age=days)`, and `age` is described as being “[measured from the resource's creation time](https://cloud.google.com/storage/docs/lifecycle#age)”.


jmarshall:
Trivially updated the commit to observe how the actions workflows are doing today…

illusional (approved):

PR #147, opened 2023-08-04
Authenticate GitHub access in actions workflows

jmarshall:
Akin to populationgenomics/cpg-infrastructure-private#182, this sets up `git+https://` authentication credentials via a credential helper in each of the workflows that needs to pip install metamist-infrastructure, which is done by checking out the relevant code from the sample-metadata repository — which is currently private so accessing it requires credentials.

See the cpg-infrastructure-private PR for further details of the technique used.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -9,9 +9,9 @@ jobs:
         shell: bash -l {0}
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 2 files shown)


jmarshall:
Updated to use the SAMPLE_METADATA_REPO_TOKEN secret name, which reflects that this token is a [fine-grained one](https://github.blog/2022-10-18-introducing-fine-grained-personal-access-tokens-for-github/) that provides access only to the sample-metadata repository.

@illusional ✈️ 🌴 Please have quick look at this one too from your 🏖️ — like the other one, the PR checks demonstrate it working.

illusional (approved):
LGTM, thanks @jmarshall!

PR #153, opened 2023-09-18
Changes to SampleMetadata config class.

milo-hyben:
Before we can deploy the latest Metamist ETL, we would need to add extra config to SampleMetadata class, here are my suggestions.
All are prefixed as ETL_

### DIFF
--- a/cpg_infra/config/config.py
+++ b/cpg_infra/config/config.py
@@ -124,7 +124,7 @@ class CPGInfrastructureConfig(DeserializableDataclass):
         slack_channel: str | None = None
         etl_accessors: list[str] = dataclasses.field(default_factory=list)
         # Metamist environment (DEVELOPMENT / PRODUCTION) for ETL cloud functions
-        etl_environment: str | None = None
+        etl_environment: str | None = 'PRODUCTION'
         # Default ETL parser configuration, if not specified in ETL payload
         # e.g.: {'project': 'greek-myth', 'default_sequencing_type': 'genome'}
         etl_parser_default_config: dict[str, str] | None = None
--- a/cpg_infra/config/config.py
+++ b/cpg_infra/config/config.py
@@ -126,6 +126,7 @@ class CPGInfrastructureConfig(DeserializableDataclass):
         # Metamist environment (DEVELOPMENT / PRODUCTION) for ETL cloud functions
         etl_environment: str | None = None
         # Default ETL parser configuration, if not specified in ETL payload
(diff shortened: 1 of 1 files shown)


cpg_infra/config/config.py
```diff
@@ -123,6 +123,10 @@ class GCP(DeserializableDataclass):
         gcp: GCP
         slack_channel: str | None = None
         etl_accessors: list[str] = dataclasses.field(default_factory=list)
+        etl_environment: str | None = None
+        etl_default_config: dict[str, str] | None = None
```
illusional, line 127: Can you add a couple of comments describing what values you'd expect in these fields?

cpg_infra/config/config.py
```diff
@@ -123,6 +123,10 @@ class GCP(DeserializableDataclass):
...
+        etl_private_repo_url: str | None = None
```
illusional, line 128: The private_repo_url you should be able to pull directly from the resource created in #152 (I hope)

```suggestion
```

cpg_infra/config/config.py
```diff
@@ -123,6 +123,10 @@ class GCP(DeserializableDataclass):
...
+        etl_private_repo_packages: str | None = None
```
illusional, line 129: Is this a comma separated list of packages to install, what would you expect here?

cpg_infra/config/config.py
milo-hyben: Comments and example values added, I had renamed config to 'etl_parser_default_config' to make the name clearer.

cpg_infra/config/config.py
milo-hyben: Good point, I had removed that entry.

cpg_infra/config/config.py
milo-hyben: Change the packages to a list, more clearer this way.

cpg_infra/config/config.py
```diff
@@ -123,6 +123,13 @@ class GCP(DeserializableDataclass):
         gcp: GCP
         slack_channel: str | None = None
         etl_accessors: list[str] = dataclasses.field(default_factory=list)
+        # Metamist environment (DEVELOPMENT / PRODUCTION) for ETL cloud functions
+        etl_environment: str | None = None
```
illusional, line 127: Can you provide 'PRODUCTION' as an explicit default here?

illusional (approved):

cpg_infra/config/config.py
milo-hyben: No worries, changed that default.

PR #154, opened 2023-09-19
Revert "Add python artifact registry"

illusional:
Reverts populationgenomics/cpg-infrastructure#152 because it keeps running a 404 error.


KatalinaBobowik (approved):
👍

PR #155, opened 2023-09-19
Artifact registry v2 (+ req bump)

illusional:
Use project_id, not the project directly to avoid the `projects/` prefix
+ resort imports


milo-hyben (approved):
Looks good to me!

PR #166, opened 2023-10-19
Write members directly to metamist

illusional:
Paired with: https://github.com/populationgenomics/metamist/pull/568


milo-hyben (approved):
Looks good to me.

PR #169, opened 2023-11-02
Add partition logic to agg

violetbrina:
Using the property _PARTITIONDATE to query our billing tables to increase efficiency.

NOTE: This WILL NOT WORK AS INTENDED until the table is partitioned. This will be imminent.
It's also worth noting that the bigquery keyword BETWEEN is date inclusive of the start and end.


illusional (approved):
Awesome! Excited for this 😀

PR #177, opened 2023-11-07
Use prop map from the past + revert usage_end_date

illusional:
As hail batch jobs can appear in the past, good to expand this range. We should make the prop map handle this case more gracefully.

### DIFF
--- a/cpg_infra/billing_aggregator/monthly_aggregate/main.py
+++ b/cpg_infra/billing_aggregator/monthly_aggregate/main.py
@@ -215,8 +215,6 @@ if __name__ == '__main__':
     logging.getLogger('urllib3').setLevel(logging.WARNING)
     event_loop = asyncio.get_event_loop()
 
-    invoice_month = '202304'
-
     from itertools import product
 
     years = [2023]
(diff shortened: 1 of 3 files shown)


violetbrina (approved):
Simple enough changes. Haven't tested the functionality myself, but the code looks ok.

cpg_infra/billing_aggregator/monthly_aggregate/main.py
```diff
@@ -240,6 +213,17 @@ def get_billing_data(invoice_month: str) -> DataFrame:
     logging.getLogger('google').setLevel(logging.WARNING)
     logging.getLogger('asyncio').setLevel(logging.ERROR)
     logging.getLogger('urllib3').setLevel(logging.WARNING)
-    event_loop = asyncio.new_event_loop()
+    event_loop = asyncio.get_event_loop()
+
+    invoice_month = '202304'
```
milo-hyben, line 218: Just as a small one, this line seems to be redundant as later on it is replaced.

milo-hyben (approved):

PR #179, opened 2023-11-14
Metamist etl configuration for parser support

illusional:
Renames sample_metadata to metamist.


milo-hyben (approved):
Looks good to me.
It seems we need to merge cpg-infra-private first so the plume preview is not failing on the old sample-meta config records.

milo-hyben (approved):

PR #182, opened 2023-11-17
Give billing.coordinator_machine_account billing.viewer role

milo-hyben:
Give billing.coordinator_machine_account billing.viewer role to be able to interrogate projects' budget and update the BQ table.


illusional (approved):
I wonder how this will translate to Azure, but we can work that out when we have to. As soon as linting passes, happy for this.

PR #184, opened 2023-11-23
Refactor key generation logic in hail and seqr

illusional:
To be unique from 2023-01-01.
I love copilot's suggestion for the comment: _Later Michael here, I've changed my mind._

### DIFF
--- a/cpg_infra/billing_aggregator/aggregate/hail.py
+++ b/cpg_infra/billing_aggregator/aggregate/hail.py
@@ -118,8 +118,12 @@ def get_finalised_entries_for_batch(batch: dict) -> List[Dict]:
             # with changing the resource_id again, we'll only use the batch_id + job_id
             # as the key as it's sensible for us to assume that all the entries exist if
             # one of the entries exists.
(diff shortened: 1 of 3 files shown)


violetbrina (approved):

cpg_infra/billing_aggregator/aggregate/hail.py
```diff
@@ -62,6 +62,26 @@ def get_billing_projects():
...
+def infer_batch_namespace(batch: dict) -> str:
+    """
+    Infer the namespace from the batch attributes
+    """
+    namespace = batch.get('attributes', {}).get('namespace')
+    user = batch.get('user')
+    default = None
+    if namespace:
+        return namespace
+    elif user:
```
milo-hyben, line 74: elif -> if 
would be enough here, we already exit the function on previous "if" above.

violetbrina: Changed!

milo-hyben (approved):

PR #185, opened 2023-11-23
Add internal users to all internal batch projects

illusional:
Context: https://docs.google.com/document/d/1kUfezhIAsVPFEHEVngvPezfae7XORoYCjjqdW5meYNw/edit

Tested that this works!

### DIFF
--- a/cpg_infra/driver.py
+++ b/cpg_infra/driver.py
@@ -302,8 +302,14 @@ class CPGInfrastructure:
             cloud_dataset.main()
 
     def setup_hail_batch_billing_project_members(self):
+        internal_users = [
+            user
+            for user in self.config.users.values()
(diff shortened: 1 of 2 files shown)


cpg_infra/driver.py
```diff
@@ -331,11 +337,18 @@ def setup_hail_batch_billing_project_members(self):
                 _group_members = self.group_provider.resolve_group_members(
                     dataset_cloud_infra.analysis_group
                 )
-                hail_batch_username = [
+                hail_batch_usernames = {
```
milo-hyben, line 340: This is casted to a set later in func "_add_member_to_billing_project", wonder if it would be better to construct it as set at first place instead of dict and then set(), set.update() on line 346 should work as is.

milo-hyben (approved):

PR #186, opened 2023-11-27
Persist metamist members to state, so we stop rerunning

illusional:
We're rerunning that metamist call a lot of times because it's not persisting the members into state, to check on next load.
The members are ordered earlier, so list comparison is fine here.


milo-hyben (approved):
Looks fine to me.

PR #190, opened 2024-01-05
Improve documentation

illusional:
And:

- moved some configuration options around because it makes more sense,
- allow the specification of Literals in the config (that check the value)

@jmarshall, I think you're fairly fresh to this repo, so a good target to see how the explanation lands. Not saying this is complete, it can be improved, but keen to get progressive improvements out.


illusional:
@jmarshall, I'm expecting the preview to fail because I've moved a config option. I've added another PR to fix it

cpg_infra/config/deserializabledataclass.py
```diff
@@ -168,6 +168,12 @@ def try_parse_value_as_type(value: Any, dtype: Any) -> Any:
             )
         return tuple(try_parse_value_as_type(v, t) for v, t in zip(value, tuple_types))
 
+    if str(getattr(dtype, '__origin__', None)) == 'typing.Literal':
```
jmarshall, line 171: Is there a reason for not doing this as `if get_origin(dtype) is Literal`?

jmarshall (approved):
Some grammar trivia and one query about the code. Once the latter is addressed, this LGTM.

jmarshall:
> I'm expecting the preview to fail because I've moved a config option. I've added another PR to fix it

Hmmm… preview seems to have failed because _cpg-infra/README.md_ is gone but _setup.py_ wants to read the long description from it. Is this what you were expecting? Not sure I spotted the other PR, and it sounds like it was addressing something else…

cpg_infra/config/deserializabledataclass.py
illusional: Is because I'm dumb valid?

jmarshall: No, but I'd accept "temporarily forgot about that API" 😄

PR #191, opened 2024-01-08
Add job name to hail batch billing

illusional:
Don't ask me why I explicitly ignored it 🤷


milo-hyben (approved):
Looks good! Thanks for the quick fix!

Issue #148, opened 2023-08-17
Don't allow full service accounts to remove bucket


illusional:
Currently a `full` service account can delete a bucket in GCP, if the output path isn't constructed correctly. This is usually fairly painful to restore, usually requiring a pulumi delete and then reup.

Noting that storage should have access to delete any blob in the buckets, but not delete the bucket itself.

This is where we allocate the permission for the `full` group in the `gcp` abstraction: https://github.com/populationgenomics/cpg-infrastructure/blob/06b6680e89e0e26ccc2b4dfdf46056da3e92d16a/cpg_infra/abstraction/gcp.py#L417-L418

I believe it's the difference between `storage.admin` and `storage.objectAdmin`.

| **storage.admin**              | **storage.objectsAdmin**       |
|-----------------------------------|-----------------------------------|
|   firebase.projects.get           |                                   |
|   orgpolicy.policy.get1           |   orgpolicy.policy.get1           |
|   resourcemanager.projects.get2   |   resourcemanager.projects.get2   |
|   resourcemanager.projects.list2  |   resourcemanager.projects.list2  |
|   **storage.buckets.***         |                                   |
|   storage.objects.*               |   storage.objects.*               |
|   storage.multipartUploads.*      |   storage.multipartUploads.*      |

PR #194, opened 2024-01-30
Disallow deletion of GCP buckets for MUTATE bucket membership type (part 1)

dancoates:
#148 

previously a `BucketMembership.MUTATE` bucket membership would assign the storage.admin role which has elevated privileges including allowing deletion of the bucket. Changing to `storage.objectAdmin` reduces these privileges to only allow manipulation of the bucket contents but not the bucket itself.

The roles differ by `storage.admin` having the following wildcard entries:

```
storage.buckets.*
storage.managedFolders.*
```

whereas `storage.objectAdmin` has:

```
storage.managedFolders.create
storage.managedFolders.delete
storage.managedFolders.get
storage.managedFolders.list
```

and no `storage.buckets` permissions. All other permissions are the same between the roles.

So in effect this change removes the following permissions from the bucket mutate membership:

```
storage.buckets.create
storage.buckets.createTagBinding
storage.buckets.delete
storage.buckets.deleteTagBinding
storage.buckets.enableObjectRetention
storage.buckets.get
storage.buckets.getIamPolicy
storage.buckets.getObjectInsights
storage.buckets.list
storage.buckets.listEffectiveTags
storage.buckets.listTagBindings
storage.buckets.setIamPolicy
storage.buckets.update
```

### DIFF
--- a/cpg_infra/abstraction/gcp.py
+++ b/cpg_infra/abstraction/gcp.py
@@ -5,7 +5,7 @@ GCP implementation for abstract infrastructure
 import base64
 from datetime import date
 from functools import cached_property
-from typing import Any
+from typing import Any, NamedTuple
 
 import pulumi
 import pulumi_gcp as gcp
@@ -23,6 +23,11 @@ from cpg_infra.abstraction.base import (
 )
 from cpg_infra.abstraction.google_group_settings import GoogleGroupSettings
 from cpg_infra.config import CPGDatasetConfig, CPGInfrastructureConfig
+
+
+class BucketMembershipRole(NamedTuple):
+    role: str
+    resource_key: str
 
 
 class GcpInfrastructure(CloudInfraBase):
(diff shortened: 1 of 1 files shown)


dancoates:
@illusional you can see above in the diff which projects this will alter the bucket role for. My only concern is whether any of those are using any of the other permissions that get removed by this role change, particularly things like `storage.buckets.get` and `storage.buckets.getIamPolicy` which are fairly innocuous and could potentially be used somewhere. I'm not sure the best way to check if these are being used though

illusional:
Thanks for putting this together @dancoates! I wonder if we should add another BucketMembership type to our abstraction that recovers some of those permissions, like list, get, getIamPolicy. 

I think this is actually fine, I don't think anyone (or service account) actively looks through the iam policies, nor do we want them to be able to?

I think this permission will be applied in an awkward way, potentially by removing the permission, and then adding it, which could break our permissions entirely during the deploy. I guess there are a few options, we wait till there's basically no analysis happening, or we try to do this in two stages (add an extra, then delete). What do you think?

dancoates:
@illusional This has got a bit more involved than I'd like but I think it is important to have the resource key explicitly defined per bucket role so that when we remove the `storage.admin` role we can ensure that the new `StorageObjectAndBucketMutator` role maintains its same resource key and is not deleted and recreated. 

I also didn't end up allowing specifying multiple bucket membership types as I'm not sure if there's much to be gained by stacking memberships - I think each role is a superset of the previous, so allowing selecting multiple wouldn't have much effect. I'm inferring that from the role names though, I don't think I can see the policies on the roles.

illusional (approved):
Awesome work, thanks @dancoates!
Very excited to protect our buckets now ⚔️

PR #195, opened 2024-02-01
Disallow deletion of GCP buckets for MUTATE bucket membership type (part 2)

### LINKED ISSUE #148: Don't allow full service accounts to remove bucket

dancoates:
Second part to #194 

Closes #148

### DIFF
--- a/cpg_infra/abstraction/gcp.py
+++ b/cpg_infra/abstraction/gcp.py
@@ -429,10 +429,6 @@ class GcpInfrastructure(CloudInfraBase):
     ):
         if membership == BucketMembership.MUTATE:
             return [
-                # @deprecated, storage.admin is too broad of a permission as it
-                # allows for deletion of the bucket. This will be removed in a second
-                # stage of this deploy
-                BucketMembershipRole('roles/storage.admin', resource_key),
                 # This role allows mutation of bucket objects but not deletion of the
                 # bucket itself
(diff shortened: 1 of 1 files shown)


illusional (approved):
Looks great!!

PR #196, opened 2024-02-02
Improve handling of invoice month edges

illusional:
Some small suggestions based on reloading data the other day

### DIFF
--- a/cpg_infra/billing_aggregator/aggregate/utils.py
+++ b/cpg_infra/billing_aggregator/aggregate/utils.py
@@ -612,14 +612,11 @@ def upsert_rows_into_bigquery(
     able to take an arbitrary amount of rows.
     """
     if not objs:
+        logger.info('Not inserting any rows')
         return 0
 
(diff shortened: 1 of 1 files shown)


cpg_infra/billing_aggregator/aggregate/utils.py
```diff
@@ -611,6 +611,8 @@ def upsert_rows_into_bigquery(
     It has some optimisations about max insert size, so this should be
     able to take an arbitrary amount of rows.
     """
+    if not objs:
+        return 0
```
milo-hyben, line 615: There is the same check few lines down, I guess you want to remove that one.

milo-hyben (approved):

cpg_infra/billing_aggregator/aggregate/utils.py
illusional: Ah yep, good call - thanks :)

milo-hyben (approved):

PR #198, opened 2024-02-05
Billing new aggregate table

milo-hyben:
This PR include the following changes:

**- Change to billing aggregate table schema:**
Labels and SystemLabels are formatted as STRING instead of array of struct.

**- Update to AlertPolicyConditionArgs for aggregate cloud function**
Trying to fix the issue with not sending slack notification when error occurs. Added resource.type=cloud_run_revision to check as last errors contain that label.

**- New setup for aggregate table**
At the moment creating of aggregate table is not present in the cpg-infra, it looks it was created manually in the past. There is a schema for the fields, which is used when bulk insert data.

**- New setup for materialized views**
There two views used by metamist billing API to show daily billing charts and batch job details.

NOTES / REASONS:
Converted labels to STRING, instead of using STRUCT or pure JSON object will allow Labels in Materialised views (MV) so when querying for a particular job we can reuse MV instead of looking into the raw aggregate table which is large.
MVs do not support aggregation by JSON or STRUCT.

A bit of background why we need this:

Originally we used MV to aggregate data by day and by some of the keys (project, topic, etc) to show daily charts.
This worked nicely until we needed drilling into the batch jobs and extract some extra labels.

We have tried MV aggregated by batch job, but to extract labels such as jobId and its details we had to look back at the aggregate table. With small jobs this was not an issue, until we start testing with more AR runs, esp. there was one Cromwell task, which contains task running over a few days. Then we had Hail Batch containing over 8000 jobs.
Querying aggregate table to get more details for those jobs took over 15 minutes and returned over 280K records, which caused metamist API to run out of memory when processing.


milo-hyben:
Hey Michael (@illusional), it took me a while to get all this sorted. I can't run pulumi preview locally and using push to test each change is a bit hairy process.
Pulumi AI gave me some suggestions how to write some BQ specific code, but it did not work when run pulumi preview via GitHub action.
I have ended up writing a simple test script to deploy BillingAggregator plugin into my own project. After plenty of iterations locally I manage to get it working under my project.
Before we deploy I would need to rename existing aggregate table and then reinsert reformatted data back.
Rest should work.
Please review.
Thank you!

cpg_infra/billing_aggregator/aggregate/seqr.py
```diff
@@ -425,12 +425,17 @@ def generator() -> Generator[dict, None, None]:
             _obj_param_map = seqr_wide_param_map
             if 'dataset' in labels:
                 # specific override where 'dataset' is specified in GCP resource
+                # MiloH question: labels seems to have format {'key': 'KEY1, 'value': 'VAL'}
+                # so how does this work?
                 _obj_param_map = {labels['dataset']: (1.0, 1)}
```
illusional, line 430: I think it never works 😓 
Can you do the format earlier so this block works?

cpg_infra/billing_aggregator/aggregate/utils.py
```diff
@@ -1151,3 +1150,20 @@ def infer_batch_namespace(batch: dict) -> str:
             return 'main'
 
     return default
+
+
+def format_as_string(data: list[dict[str, Any]]) -> str:
```
illusional, line 1155: Can you rename this to something bigquery related?
Maybe even doing this in a two step (convert from the [{key: key, value: value}] to {} so it's easier to work with, then to string as a later when we're about to send it to bigquery.

illusional (commented):
Looks good to me!
How do we want to deploy this? We could do it over zoom locally, as:

> Note: On newer versions of the provider, you must explicitly set deletion_protection=false (and run pulumi update to write the field to state) in order to destroy an instance. It is recommended to not set this field (or set it to true) until you’re ready to destroy.

cpg_infra/billing_aggregator/aggregate/utils.py
milo-hyben: Have renamed to reformat_bigqquery_labels and keep results to dict, make changes so when we assign to BQ record it gets converted to string."

cpg_infra/billing_aggregator/aggregate/seqr.py
milo-hyben: As suggested, reformatted labels prior this check.

illusional (approved):

nevoodoo (approved):
Great work :)

PR #201, opened 2024-02-15
Increase RAM for hail aggregate function (from 2GB to 2.5GB)

milo-hyben:


illusional:
Closing in favour of #204

PR #202, opened 2024-02-15
Add read access for metamist to list objects and object metadata in dataset buckets 

dancoates:


illusional (approved):
Awesome, looks good to me!

PR #204, opened 2024-02-15
Improve memory efficiency of billing aggregators

illusional:
Rather than:

1. getting all batches, then
2. getting all jobs for chunks of batches, then
3. process the entries, inserting early when size of entries is high

Let's change some of this to yield entries back, and process on the fly:

1. Get all batches, then
2. For chunks of batches, trigger off the get_jobs call
3. Grab each job as it's loaded and add to a queue,
4. Process jobs as we get them, from any batch, at any time
5. Insert them all at once, or whenever we reach 10MB (but this is configurable)

### DIFF
--- a/cpg_infra/billing_aggregator/aggregate/hail.py
+++ b/cpg_infra/billing_aggregator/aggregate/hail.py
@@ -26,7 +26,7 @@ import logging
 import os
 import shutil
 from datetime import datetime
-from typing import Any, Dict, List
+from typing import Any, Generator
 
 import functions_framework
 from flask import Request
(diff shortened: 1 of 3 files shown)


milo-hyben (approved):
Looking good to me, will be interesting to see how that performs with the Q.

Just a small one, shouldn't this be as UTC as well:
https://github.com/populationgenomics/cpg-infrastructure/blob/b0b7fb7d9dccab23a5943c724b75c4189f60904f/cpg_infra/billing_aggregator/aggregate/seqr.py#L197

illusional:
Great catch, thank you @milo-hyben (do you mind re-approving with the latest commit)? 
We can deploy this next week, there are currently some issues with group memberships.

milo-hyben (approved):

PR #205, opened 2024-02-19
Dynamic group membership provider

dancoates:


cpg_infra/abstraction/google_group_membership.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def create(self, props: GoogleGroupMembershipProviderInputs):
+        group_id = props['group_id']
+        member_key = props['member_key']
+
+        member = find_member_in_group_by_key(group_id, member_key)
+
+        # If the group membership already exists, then don't try and create it
+        if member != None:
+            return pulumi.dynamic.CreateResult(id_=member['member_name'])
```
illusional, line 77: I think google groups use the nomenclature:

- group key, that alphanumeric ID
- member_key: the email
- member_id: an alphanumeric ID

If you store all three on the CreateResult, you can get them all as part of the list.

cpg_infra/abstraction/google_group_membership.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def find_member_in_group_by_key(group_id: str, member_key: str):
+    member_list = list_group_members(group_id)
+    return next((m for m in member_list if m['member_key'] == member_key), None)
+
+
+def find_member_in_group_by_name(group_id: str, member_name: str):
+    member_list = list_group_members(group_id)
+    return next((m for m in member_list if m['member_name'] == member_name), None)
```
illusional, line 116: I don't _love_ this find by iteration. If it's possible to return a dict from above, and check presence of the member_key in all situations that's my preferred. It's not a big deal, but this would could get called like 10-20k times on a refresh, and some of the group sizes can get quite large.

cpg_infra/abstraction/google_group_membership.py
```diff
@@ -0,0 +1,207 @@
...
+        # If the member has already been removed then no need to remove
+        if member == None:
+            return
+
+        remove_member_from_group(id)
+
+    def diff(self, id: str, olds: Any, news: Any):
+        # There isn't anything to change on a group membership, they are either a member
+        # or not, so this can always return false
+        return pulumi.dynamic.DiffResult(changes=False)
```
illusional, line 98: I'm pretty sure, if you throw different params but the same key, I think pulumi will run this `diff`. And a different member_key should always cause a recreate with `replaces`: https://www.pulumi.com/docs/concepts/resources/dynamic-providers/#diffid-olds-news

illusional (approved):

illusional (approved):

nevoodoo (approved):

PR #208, opened 2024-02-27
Infra changes to allow a development deploy

dancoates:


cpg_infra/abstraction/hailbatch.py
```diff
@@ -31,7 +31,17 @@
 )
 
 
-def get_hail_batch_token(token_category) -> str:
+def get_hail_batch_auth_headers(token_category: str, batch_uri: str) -> dict[str, str]:
+    token = get_hail_batch_auth_token(token_category)
+    headers = {'Authorization': f'Bearer {token}'}
+    # If this is a dev service then need an extra header
+    if batch_uri.startswith('https://internal.hail'):
+        headers['X-Hail-Internal-Authorization'] = f'Bearer {token}'
```
illusional, line 39: Does this work? My understanding was these are actually different auth tokens [team-docs/hail.md](https://github.com/populationgenomics/team-docs/blob/main/hail.md#:~:text=This%20now%20requires%20two%20tokens%3A%20one%20for%20the%20default%20namespace%2C%20and%20the%20other%20for%20your%20development%20namespace.%20They%20need%20to%20be%20passed%20as%20separate%20headers%3A)?

cpg_infra/config/config.py
```diff
@@ -205,6 +213,13 @@ class GCPAggregator(DeserializableDataclass):
     # within itself, but this is a map of all users known to the system
     users: dict[MemberKey, CPGInfrastructureUser]
 
+    # include list of plugins enabled. This is specified explicitly to control
+    # what plugins are included from dependencies, also to allow exclusion of
+    # plugins for testing purposes
+    plugins_enabled: list[str] = dataclasses.field(
+        default_factory=lambda: ['metamist', 'billing_aggregator'],
```
illusional, line 220: Happy for this to be always specific in the config rather than the two plugins being the default. I like this though, if a plugin isn't installed, it will fail.

cpg_infra/config/config.py
```diff
@@ -205,6 +213,13 @@ class GCPAggregator(DeserializableDataclass):
     # within itself, but this is a map of all users known to the system
     users: dict[MemberKey, CPGInfrastructureUser]
 
+    # include list of plugins enabled. This is specified explicitly to control
+    # what plugins are included from dependencies, also to allow exclusion of
+    # plugins for testing purposes
```
illusional, line 218: Can you add a small comment just noting they're exposed through entrypoints, and to check <file> for more documentation?

cpg_infra/driver.py
```diff
@@ -295,19 +295,42 @@ def resolve_dataset_order(self):
...
 
+        # Deploy all the assets required for each dataset. Groups, permissions
+        # storage buckets, metamist and hail users etc.
         self.deploy_datasets()
 
-        for plugin in get_plugins().values():
-            plugin(self, self.config).main()
+        for plugin_name, plugin in get_plugins().items():
+            if plugin_name in self.config.plugins_enabled:
+                plugin(self, self.config).main()
```
illusional, line 321: Following on from my comment before, I think loop through the plugins, and if one isn't found raise an Exception.

illusional (approved):
Looking great! This repo has _last commit requires review_ - so I'll need to re-approve after feedback is made :)
Can you also expand out the PR description here to provide more context for those following along at home?

cpg_infra/abstraction/hailbatch.py
dancoates: Yeah it does work, I think because I'm using GCP credentials the same credentials work for dev and prod. I'm not sure if there's another type of token that would require different ones for dev and prod?

cpg_infra/config/config.py
dancoates: I made it the default mostly so that the linting on this branch would pass. I can definitely remove but we'll need to merge the private PR first or this one will fail when it tries to preview because of the missing config.

dancoates: I've removed the default value, we can just co-ordinate the merges to do the private one first

cpg_infra/abstraction/hailbatch.py
illusional: Cool! There are session tokens you can provide (the one you created and put in tmp) and supply those. That's actually how the billing aggregator works in prod (there's an environment variable) for each environment. 

Not to say we shouldn't move that over to GCP / Azure credentials to do that :) [but maybe for later]

illusional (approved):

illusional (approved):

PR #210, opened 2024-03-06
Automatic hail user creation

dancoates:
Removes the last manual step of creating a new stack

### DIFF
--- a/cpg_infra/config/config.py
+++ b/cpg_infra/config/config.py
@@ -91,7 +91,7 @@ class CPGInfrastructureConfig(DeserializableDataclass):
 
         gcp: GCP
         azure: Azure | None = None
-        username_prefix: str | None = ''
+        username_prefix: str | None = None
 
     @dataclasses.dataclass(frozen=True)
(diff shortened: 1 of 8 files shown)


cpg_infra/config/config.py
```diff
@@ -79,16 +79,19 @@ class Hail(DeserializableDataclass):
...
             wheel_bucket_name: str | None = None
 
         @dataclasses.dataclass(frozen=True)
         class Azure(DeserializableDataclass):
             hail_batch_url: str
+            hail_auth_url: str
 
         gcp: GCP
         azure: Azure | None = None
+        username_prefix: str | None = ''
```
illusional, line 94: Should this default to `None` or not allow None?

```suggestion
        username_prefix: str | None = None
```

illusional (approved):

cpg_infra/config/config.py
dancoates: good pickup, I wasn't handling None's properly for the username prefix, have fixed that now

illusional (approved):

PR #211, opened 2024-03-12
Remove isort + fix pyproject to sort

illusional:
There's some ambiguity with graphlib, as it moved from external to internal in 3.11, but Ruff doesn't support the isort py_version argument. Broadly I think this is actually okay, I'm not fussed with it's placement, as long as it's not inconsistently moved.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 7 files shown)


dancoates (approved):

dancoates:
Looks good, but linting is failing at the mo, not sure if just a glitch or if something is inconsistent between local and CI environment?

dancoates (approved):

PR #213, opened 2024-04-01
Disable pulumi preview

illusional:
On public branches


dancoates (approved):
Looks good, you've probably already done it, but if not would be good to remove the `INFRA_PRIVATE_DEPLOY_TOKEN` env var from the workflow config in github too

PR #214, opened 2024-04-03
Use new production-yaml path

illusional:
Changed in: https://github.com/populationgenomics/cpg-infrastructure-private/pull/290


dancoates (approved):
huh, I thought I checked everywhere for instances of `production.yaml`, must have missed this one!

PR #215, opened 2024-04-05
Aggregate changes for large batches

milo-hyben:
This PR contains:
1. Small materialised view change to be bale to see hail batch query jobs
2. Disable monthly aggregate function to populate google doc as this has been replaced with metamist billing report.
3. Aggregate large batches (over 9K jobs per batch - this is set by DEFAULT_MAX_JOBS_PER_BATCH), all jobs billing information gets merged into one job with the JobId of the last jobs, this help us to keep number of jobs per batch.
4. Increase cloud function from 9mins to 1H, there is only one concurrent allowed so if Subscription triggers another after 10mins (max PubSub timeout possible for push), it would fail and wait for the first to complete.

I have tested on Feb 21-Feb 20, which took over 4 hours before to run from local machine. Much faster now.
We can discuss if 9K is not too restrictive, but noticed not many non hail batch query have more than 8K of jobs.


cpg_infra/billing_aggregator/aggregate/utils.py
```diff
@@ -1223,3 +1238,60 @@ def reformat_bigqquery_labels(data: list[dict[str, Any]]) -> dict[str, Any]:
...
+        result = get_bigquery_client().query(_query, job_config=job_config).result()
+        records = set(result.to_dataframe()['id'])
+    except Exception as e:  # noqa: BLE001
+        logger.error(e)
+
+    logger.info(
+        f'Retrieved {len(records)} stored ids',
+    )
+
+    return records
```
illusional, line 1297: Why wouldn't we want to FAIL if we can't get the records, and return an empty set - this would mean duplicate records

cpg_infra/billing_aggregator/driver.py
```diff
@@ -62,7 +62,7 @@ def main(self):
             return
 
         self.setup_aggregator_functions()
-        self.setup_monthly_export()
+        # self.setup_monthly_export() # we do not need this anymore
```
illusional, line 65: Can you remove all the code for this function to?

```suggestion
```

cpg_infra/billing_aggregator/aggregate/utils.py
```diff
@@ -464,6 +469,7 @@ async def get_finished_batches_for_date(
 async def get_jobs_for_batch(
     batch_id: int,
     token: str,
+    limit: int | None = None,
```
illusional, line 472: These two limits are for different things, this limit is the _number of jobs to get per request_, not the total number of jobs before we back out. I think you're better to do the query_on_batch check before it gets to here, even write a specific function that takes in a batch_record, and returns True if we guess it to be a hail query on batch, batch (name is blank, and n_jobs > 10k ish).

illusional (changes requested):
Looking good! We did a heap of optimisation so that we don't need the staged ranges (ie: timedelta(day=1)), I'm not opposed to re-adding it for a debug mode, but I think it might be better to be build into the "process_start_end" to produce a list of date ranges [ like it used to be before the optimisation ].

I think the max job limit for detecting hail query isn't quite right, happy to chat if I'm misunderstanding!

cpg_infra/billing_aggregator/aggregate/utils.py
milo-hyben: Yes, makes sense, removed the try/except.

cpg_infra/billing_aggregator/driver.py
milo-hyben: Done, removed bot call and the function.

cpg_infra/billing_aggregator/aggregate/hail.py
```diff
@@ -226,6 +227,29 @@ async def main(
     return {'entriesInserted': result}
 
 
+def reload_data_example():
+    """Re-loading one month example call"""
+    test_start = datetime.fromisoformat('2024-02-01')
+    test_end = datetime.fromisoformat('2024-02-28')
+
+    current_date = test_start
+    while current_date <= test_end:
```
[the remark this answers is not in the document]
milo-hyben: As function name suggests *_example() it is just example how to load day by day. I had only used it to load multiple days in a row for testing. 
The __main__ function is example/test as well it is just doing from now to 4 hours prior.
I had removed it.

cpg_infra/billing_aggregator/aggregate/utils.py
milo-hyben: Removed that, taking care of batch queries prior, checking name attribute.

milo-hyben:
Made a few changes, reflecting your comments:
1. Hail Batch Query is identified now as batch with no attribute name
2. Still cut max jobs per batch if jobs_cnt > 9K (happy to discuss if too restrictive)
3. Added option to load batches by batch_id

Please let me know what do you think.

cpg_infra/billing_aggregator/aggregate/utils.py
```diff
@@ -599,35 +624,34 @@ async def _get_jobs_and_add_to_queue(
...
         jobs_cnt = batch['n_jobs']
-
-        # only get the first jb if there are more than DEFAULT_MAX_JOBS_PER_BATCH jobs
-        limit = 1 if jobs_cnt > DEFAULT_MAX_JOBS_PER_BATCH else None
-        async for jobs in get_jobs_for_batch(batch_id, token, limit):
-            if jobs_cnt <= DEFAULT_MAX_JOBS_PER_BATCH:
+        batch_name = batch.get('attributes', {}).get('name')
+
+        # check if hail batch query or too many jobs
+        if batch_name is None or jobs_cnt > DEFAULT_MAX_JOBS_PER_BATCH:
```
illusional, line 630: - We have batches over the max_jobs_per_batch that we want to process
- We have batches without names under the count we want to process

```suggestion
        if batch_name is None and jobs_cnt > DEFAULT_MAX_JOBS_PER_BATCH:
```

cpg_infra/billing_aggregator/aggregate/hail.py
```diff
@@ -190,6 +195,25 @@ def from_request(request: Request):
...
+    for batch_id in batch_ids:
+        batch = await utils.get_batch_by_id(batch_id, token=utils.get_hail_token())
+        start_time = utils.parse_hail_time(batch['time_created'])
+        end_time = utils.parse_hail_time(batch['time_completed'])
+        if start is None or start_time < start:
+            start = start_time
+        if end is None or end_time > end:
+            end = end_time
+
+    return await main(start, end, batch_ids=batch_ids)
```
illusional, line 214: The start / end are not used batch_ids flow of `main` - so I don't think there's a need for this.
And if for some reason you do, set them in the `main` function where you get the batches.

illusional (approved):
Looks good! Two changes that must be addressed before merging, always happy to re-review if you want to double check my suggestions

cpg_infra/billing_aggregator/aggregate/utils.py
milo-hyben: Ok, that makes sense, I have thought we want to skip very large ones no matter if HB Query or not. Will change to 'and' so only very large HB Query will be aggregated.

cpg_infra/billing_aggregator/aggregate/hail.py
milo-hyben: Yes, we need the start and end date to retrieve 'stored_ids' from BQ table, prior to calling each batch separately. This was intruded in this PR to optimise those multiple BQ calls with each Hail Batch individually.

milo-hyben:
I had added CI Jobs and for jobs with 0 cost, it adds 0 cost to job resources so it is included in the aggregate table.
I had checked Oct-Dec 2023 and we are missing 1306 jobs, will slack you more details.

cpg_infra/billing_aggregator/aggregate/utils.py
```diff
@@ -513,24 +513,46 @@ async def get_jobs_for_batch(
...
     token: str,
+    attempts: int = 5,
 ) -> BatchType:
     """
     Get batch details by batch_id
     """
     try:
         url = HAIL_BATCH_API.replace('{batch_id}', batch_id)
-        logger.info(f'Getting batch: {url}')
+        # logger.info(f'Getting batch: {url}')
```
illusional, line 523: You don't like this logging?

cpg_infra/billing_aggregator/aggregate/utils.py
```diff
@@ -624,10 +650,28 @@ async def _get_jobs_and_add_to_queue(
...
+            # with the rest of the code
+            empty_job = {
+                'batch_id': batch_id,
+                'job_id': jobs_cnt,
+                'state': batch.get('state'),
+                'resources': {},
+                'attributes': {},
+                'cost': {
+                    # we need some entry in the cost, so we can insert it
+                    'ip-fee/1024/1': 0,
```
illusional, line 665: Do you want to call this something notable so it's easier to track down in the future, like '<empty-batch-cost-placeholder>' or will that break things?

cpg_infra/billing_aggregator/aggregate/utils.py
```diff
@@ -1083,6 +1130,41 @@ def get_batch_ids_from_request(
     return None
 
 
+def get_batch_ids_from_data(
+    data: str | dict | None,
+) -> list[str] | None:
+    """
+    Get batch_id from the cloud function data.
+    """
+    if data is not None:
```
illusional, line 1139: Prefer to early-return in Python if practical

```python
if data is None:
    return None

# do other stuff
```

cpg_infra/billing_aggregator/aggregate/utils.py
```diff
@@ -513,24 +513,46 @@ async def get_jobs_for_batch(
 async def get_batch_by_id(
```
illusional, line 513: Ehhh, it's not a big deal I guess, but I see lots of individual calls to this. I wonder if it's worth adding a cache decorator to this function?

illusional (approved):
Looking good!

cpg_infra/billing_aggregator/aggregate/utils.py
milo-hyben: Was printing out missing batches, so disabled it for my testing. I will put back in.

cpg_infra/billing_aggregator/aggregate/utils.py
milo-hyben: Let me try '', makes more sense.

cpg_infra/billing_aggregator/aggregate/utils.py
milo-hyben: It is called with different params, so do not think it is wort it. It would be called max 2x per run: 
if loaded by batch id (max 2x), if by date then only once.

cpg_infra/billing_aggregator/aggregate/utils.py
milo-hyben: So yes {' ': 0} works.

milo-hyben:
I had resolved the comments. One last review please(can't merge without it). Thank you!

illusional (approved):

PR #221, opened 2024-04-22
Seqr - memory optimisation - only add missing records to the list of entries to be updated.

milo-hyben:
Background, when trying to process batch https://batch.hail.populationgenomics.org.au/batches/430882
, which took over a week to process, the function runs out of memory because it is trying to prepare the list of all gcp records for that time. Even most of them would be already upserted as seqr scheduled aggregation runs every 4 hours.

This PR checks if records exist and only add it to a list if not present.
This should minimise the list passed around and therefore minimising function RAM requirements.

### DIFF
--- a/cpg_infra/billing_aggregator/aggregate/seqr.py
+++ b/cpg_infra/billing_aggregator/aggregate/seqr.py
@@ -443,7 +443,8 @@ def migrate_entries_from_bq(
             obj['id'] = nid
 
             # For every seqr billing entry migrate it over
-            entries.append(obj) if obj['id'] not in existing_ids else None
+            if obj['id'] not in existing_ids:
+                entries.append(obj)
 
             # For every seqr billing entry, add credit entry
             obj_credit = utils.get_credit(
@@ -451,7 +452,8 @@ def migrate_entries_from_bq(
                 topic='seqr',
                 project=utils.SEQR_PROJECT_FIELD,
             )
-            entries.append(obj_credit) if obj_credit['id'] not in existing_ids else None
+            if obj_credit['id'] not in existing_ids:
+                entries.append(obj_credit)
 
             for dataset, (ratio, dataset_size) in _obj_param_map.items():
                 new_entry = copy.deepcopy(obj)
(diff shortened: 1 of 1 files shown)


cpg_infra/billing_aggregator/aggregate/seqr.py
```diff
@@ -395,8 +395,16 @@ def generator() -> Generator[dict, None, None]:
 
             json_objs_iter = generator()
 
+    def _append_if_not_present(entries: list[dict], obj: dict) -> None:
+        """
+        Append records if not already in the table
+        """
+        if obj['id'] not in existing_ids:
+            # only insert if it's not already in the table
+            entries.append(obj)
```
illusional, line 404: Maybe a nitpick, but I don't think you need this function, it's logically pretty simple (not complex enough for an abstraction), what do you think?

illusional (approved):

cpg_infra/billing_aggregator/aggregate/seqr.py
milo-hyben: Good point, saving 2 lines in the loop can be done differently.

cpg_infra/billing_aggregator/aggregate/seqr.py
```diff
@@ -443,14 +443,15 @@ def generator() -> Generator[dict, None, None]:
...
 
             # For every seqr billing entry migrate it over
-            entries.append(obj)
-            entries.append(
-                utils.get_credit(
-                    entry=obj,
-                    topic='seqr',
-                    project=utils.SEQR_PROJECT_FIELD,
-                ),
+            entries.append(obj) if obj['id'] not in existing_ids else None
```
illusional, line 446:
```suggestion
            if obj['id'] not in existing_ids:
                entries.append(obj)
```

milo-hyben: Yes, that one was my original, than I did not like the main loop does not fit on the screen.
But I guess it reads better.

illusional (approved):

PR #224, opened 2024-04-23
Limit hail batches to max 50K jobs, aggregate any larger ones.

milo-hyben:
When very large batches are being loaded (e.g. https://batch.hail.populationgenomics.org.au/batches/444166), cloud function was running out of memory. This particular batch had over 370K jobs, which generated over 4 million records for the billing BQ table.
We had decided to aggregate hail batches if they reach 50K jobs.

### DIFF
--- a/cpg_infra/billing_aggregator/aggregate/utils.py
+++ b/cpg_infra/billing_aggregator/aggregate/utils.py
@@ -94,10 +94,10 @@ ANALYSIS_RUNNER_PROJECT_ID = 'analysis-runner'
 
 # Maximum job count before all jobs gets summarised into one row
 # This is to prevent the bigquery inserts taking too long
-# Seqr jobs over 9K are going to be aggregated into one job.
-SEQR_MAX_JOBS_PER_BATCH = 9000
-# For hail jobs this limit is 50K
-HAIL_MAX_JOBS_PER_BATCH = 50000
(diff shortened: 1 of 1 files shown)


cpg_infra/billing_aggregator/aggregate/utils.py
```diff
@@ -671,8 +673,10 @@ async def _get_jobs_and_add_to_queue(
 
         batch_name = batch.get('attributes', {}).get('name')
         # check if hail batch query and too many jobs
-        if batch_name is None and jobs_cnt > DEFAULT_MAX_JOBS_PER_BATCH:
-            # This is most likely Hail Batch Query job, aggregate it as one job
+        if (batch_name is None and jobs_cnt > SEQR_MAX_JOBS_PER_BATCH) or (
+            batch_name is not None and jobs_cnt > HAIL_MAX_JOBS_PER_BATCH
```
illusional, line 677: I think this is right, but the names of these variables are confusing, can you align them closer to what they represent, eg: `HAIL_QUERY_JOB_PER_BATCH_LIMIT` and `HAIL_NON_QUERY_JOB_PER_BATCH_LIMIT`?

illusional (changes requested):

illusional (approved):

PR #226, opened 2024-04-24
Optimising seqr billing aggregation

milo-hyben:
1. Make sure BQ to collect existing stored is run only once
2. Increase RAM for seqr function to 4GB.


ckrishk (approved):
Looking good to me

milo-hyben:
Closing this PR without merging, it has been replaced with better solution:
https://github.com/populationgenomics/cpg-infrastructure/pull/238

PR #232, opened 2024-05-13
Make storage-viz more durable

illusional:
An intermediate storage viz (https://batch.hail.populationgenomics.org.au/batches/445677) failed for an [out-of-memory issues](https://batch.hail.populationgenomics.org.au/batches/445677/jobs/5).

There are two options, but the easiest is to bump the memory, and fail the dataset. Some later options is partition by bucket instead of dataset, but let's do it later if we need.

Also make the last aggregate bash always run, and just post an error if datasets are missing.

Also, noting the actual path construction was pretty disconnected, an output_prefix was supplied, but then the output_path function was used later to construct the web URL prefix. I tested the aggregation works locally.

### DIFF
--- a/storage_visualization/disk_usage.py
+++ b/storage_visualization/disk_usage.py
@@ -11,6 +11,7 @@ from collections import defaultdict
 import google.api_core.exceptions
 from cloudpathlib import AnyPath
 from google.cloud import storage
+from humanize import naturalsize
 
 from cpg_utils.config import get_access_level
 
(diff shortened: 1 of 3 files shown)


dancoates (approved):
These changes look good, I do wonder if there's a bit of a memory leak issue though given that this is using so much memory when it just just doing a fairly simple aggregation. I had a quick look and can't see anything obvious, will be interesting to see where the memory usage peaks with the `highmem` option

illusional:
Memory usage:
image

Chatty summary doesn't really give anything surprising:

- `blobs = storage_client.list_blobs(bucket_name)`: If there are a huge number of blobs, this can potentially consume a lot of memory since the list might be very large. Consider using pagination or streaming the blobs if supported by the library to avoid loading all blobs into memory at once.
    - Though, given it's able to report millions of blobs summarised already, I find this pretty unlikely this is the problem. Unless it's not purging them from memory?

- `stats = aggregate_stats_container[path]`: If aggregate_stats_container is a dictionary that grows indefinitely with new paths, this can lead to a memory leak as more and more entries are added. Make sure to monitor and manage the size of this container, possibly by purging old data if it's no longer needed.
    - I've added some reporting every million blobs to report the memory used - which is where I suspect it's actually being consumed.

dancoates:
Yeah, I know python dicts are pretty inefficient memory wise so it could just be the sheer amount of paths stored in the `aggregate_stats_container` dict that's causing the high memory usage.

dancoates (approved):

PR #234, opened 2024-05-16
Pinning metamist to new version, forcing metamist-infra update.

milo-hyben:
Trying to upgrade metamist-private in the cloud etl function.


illusional (approved):
Gross, we'll have to work out a better way to do this I guess. I wonder if just a manual GH action to rebuild + deploy the ETL instance? Even put it in the metamist repo

PR #237, opened 2024-05-21
Update storage vis treemap job to highmem

dancoates:


illusional (approved):
Sure, see if it uses less memory than this :) 
Otherwise keen for your proposed format changes to dramatically improve this :)

PR #238, opened 2024-06-19
Replacing Billing Aggregate Pub/Sub with Cloud Scheduler

milo-hyben:
This PR includes:
- replace Pub/Sub approach for loading billing data with Cloud Scheduler, which allows to run for up to 1 hours (Pub/Sub has 9 mins limit)
- remove the update billing budget function as this is not used and will be implemented as part of cpg-infra-private GitHub action.

### DIFF
--- a/cpg_infra/billing_aggregator/aggregate/requirements.txt
+++ b/cpg_infra/billing_aggregator/aggregate/requirements.txt
@@ -5,8 +5,9 @@ metamist
 pyyaml
 python-rapidjson
 google-api-python-client==2.79.0
-google-cloud-bigquery==3.6.0
+google-cloud-bigquery==3.11.4
 google-cloud-secret-manager==2.15.1
 google-cloud-logging==3.5.0
-pandas==1.5.3
-pandas-gbq==0.19.1
+numpy==1.23.2
+pandas==2.0.1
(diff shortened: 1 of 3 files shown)


cpg_infra/billing_aggregator/aggregate/requirements.txt
```diff
@@ -5,8 +5,9 @@ metamist
 pyyaml
 python-rapidjson
 google-api-python-client==2.79.0
-google-cloud-bigquery==3.6.0
+google-cloud-bigquery==3.11.4
 google-cloud-secret-manager==2.15.1
 google-cloud-logging==3.5.0
-pandas==1.5.3
-pandas-gbq==0.19.1
+numpy==1.23.2
```
illusional, line 11: Is numpy just pinned, or was it introduced?

illusional (approved):

cpg_infra/billing_aggregator/aggregate/requirements.txt
milo-hyben: It is pinned, I was having issue with pip install during cloud function setup with numpy 2.0.0 which is now default. This version was suggested on stackoverflow, I just can't find the reference atm.

PR #245, opened 2024-07-24
Migrate slack bot

violetbrina:
Final commit moving the source archive out of the loop wasn't commited. Getting a 'billing-aggregator-source-code' already exists error. The move should fix it.


dancoates (approved):
whoops, sorry I should have been more thorough in my re-review

PR #246, opened 2024-07-24
CPG Cost Control bot migration to cpg-infra

violetbrina:
As it says. Matching private PR [here](https://github.com/populationgenomics/cpg-infrastructure-private/pull/366).

### DIFF
--- a/cpg_infra/billing_aggregator/driver.py
+++ b/cpg_infra/billing_aggregator/driver.py
@@ -151,12 +151,11 @@ class BillingAggregator(CpgInfrastructurePlugin):
             source=archive,
             opts=pulumi.ResourceOptions(replace_on_changes=['*']),
         )
-    
-    
+
     def setup_gcp_cost_control(self):
         """
-            Create the gcp cost control cloud function to cut off billing when
-            it exceeds the budget
+        Create the gcp cost control cloud function to cut off billing when
+        it exceeds the budget
         """
         region = self.config.gcp.region
         service_account = self.config.billing.gcp_cost_control.machine_account
@@ -191,8 +190,6 @@ class BillingAggregator(CpgInfrastructurePlugin):
         )
 
         pulumi.export('gcp_cost_control_cloud_function', function)
-
-
 
     def setup_gcp_cost_reporting(self):
         """
@@ -267,7 +264,6 @@ class BillingAggregator(CpgInfrastructurePlugin):
 
         pulumi.export('gcp_cost_reporting_cloud_function', function)
         pulumi.export('gcp_cost_reporting_cloud_scheduler_job', cron_job)
-
 
     def setup_aggregator_functions(self):
         """Setup hourly aggregator functions"""
--- a/cpg_infra/billing_aggregator/driver.py
+++ b/cpg_infra/billing_aggregator/driver.py
@@ -164,29 +164,37 @@ class BillingAggregator(CpgInfrastructurePlugin):
 
         # Create source archive
         source_archive = self.create_source_archive(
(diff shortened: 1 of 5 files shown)


cpg_infra/billing_aggregator/driver.py
```diff
@@ -151,6 +152,45 @@ def create_source_archive(
...
+        it exceeds the budget
+        """
+        region = self.config.gcp.region
+        service_account = self.config.billing.gcp_cost_controls.machine_account
+        slack_channel = self.config.billing.gcp_cost_controls.slack_channel
+        pubsub_topic_name = self.config.billing.gcp_cost_controls.pubsub_topic
+
+        # Create source archive
+        source_archive = self.create_source_archive(
+            'billing-gcp-cost-report-source-code',
```
dancoates, line 167: This'll need a different resource name tot he cost report bot I think?

cpg_infra/billing_aggregator/driver.py
```diff
@@ -151,6 +152,45 @@ def create_source_archive(
...
+            self.source_bucket.name,
+            os.path.join(os.path.dirname(__file__), 'gcp_cost_control'),
+        )
+
+        # Deploy Cloud Function
+        function = gcp.cloudfunctionsv2.Function(
```
dancoates, line 173: Does this work? I would have thought this would need the `build_config` and `service_config` stuff that is in `create_cloud_function`? Or is this meant to be calling `self.create_cloud_function`?

dancoates (changes requested):
Looks good, just one little thing, and one thing I'm confused by

cpg_infra/billing_aggregator/driver.py
violetbrina: Great spot. Yup changed.

cpg_infra/billing_aggregator/driver.py
violetbrina: Great pickup. You're right! I'd originally used the create_cloud_function until I realised it would be too hard to modify to create a pubsub triggered function. Forgot to wrap the build config parameters correctly.

dancoates: Interesting that our type checking / linting didn't pick anything up, I guess the gcp cloudfunctions method doesn't have typing info attached.

milo-hyben (approved):
Looking good to me, we just need to double check if topic name needs to be full path or just name itself.

dancoates (approved):

cpg_infra/billing_aggregator/driver.py
```diff
@@ -151,6 +152,45 @@ def create_source_archive(
...
+            source_file='main.py',
+            entry_point='gcp_cost_control',
+            runtime='python311',
+            available_memory_mb=256,
+            environment_variables={
+                'SLACK_CHANNEL': slack_channel,
+            },
+            event_trigger=gcp.cloudfunctionsv2.FunctionEventTriggerArgs(
+                event_type='google.pubsub.topic.publish',
+                pubsub_topic=pubsub_topic_name,
```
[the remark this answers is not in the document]
dancoates: I _think_ it might be ok, the docs for the pulumi config aren't very clear but the cli docs say that it doesn't need the full name including the project:
https://cloud.google.com/sdk/gcloud/reference/functions/deploy#--trigger-topic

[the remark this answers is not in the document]
violetbrina: Yup the pubsub_topic takes the topic name only. I believe because it needs to be in the same project/region anyway.

dancoates (approved):

PR #248, opened 2024-07-30
Adjust Billing BQ filters to use partitions

milo-hyben:
This change:
1. Optimise BQ queries, which did not use the benefit of partitions. Would lower the cost of running the queries.
2. Some of the filters were too restrict, esp. when checking existing ids by usage_end_time, this is fixing bug with some random duplicated records.
3. pin numpy to <2, as some versions of python have issue with version 2.0 as mentioned here: https://github.com/numpy/numpy/issues/26710


violetbrina (approved):
Looks good to me. I particularly like the small and large filter windows for different use cases.

milo-hyben:
I had to merge from main, @violetbrina please approve again, thank you!

violetbrina (approved):

PR #249, opened 2024-08-01
Increase RAM for Seqr aggreg billing job.

milo-hyben:
Recent runs did reach up to 2.7GB of ram, increase the RAM for seqr billing to 3GB.


illusional (approved):
I suspect it will just run out again, I wonder if we can work out why it's running out of memory, and see if there are optimisations to make there (or if it's only the IDs thing, do a count first and choose a different method for conflict detection if it's huge).

PR #253, opened 2024-08-07
Misc fixes

illusional:
- Pubsub messages in cloudrun have a slightly revised format: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/f7828705deaeb743828a531d5c25bc2cc6505a06/run/pubsub/main.py#L30-L45
- Need to include `__init__.py` in the directory so it gets carried across during the `pip install` (or included in package data)
- Bump memory for seqr
- Reduce time to look forward / backwards (@milo-hyben, worth revisiting this pre-fetch IDs approach, as it's causing so many more memory issues


violetbrina (approved):
Looks good to me.

PR #257, opened 2024-08-17
Add Test Summary to Lint + Test workflow

violetbrina:


.github/workflows/lint.yaml
```diff
@@ -39,12 +42,23 @@ jobs:
...
-        run: pre-commit run --all-files
-
       - name: Fail if tests are not passing
         if: ${{ steps.runtests.outputs.rc != 0 }}
         uses: actions/github-script@v6
         with:
           script: |
             core.setFailed('Unit tests failed with rc = ${{ steps.runtests.outputs.rc }}')
+
+      - name: Publish Test Results
```
illusional, line 52: I think it'd be great to do this before the "fail the workflow if the tests don't pass" step so you get to see the test report as a comment.

nevoodoo (approved):

.github/workflows/lint.yaml
violetbrina: Good suggestion. Moved!

illusional (approved):

PR #264, opened 2024-09-04
Adjusting billing aggregation timeouts and memory.

milo-hyben:
Increased RAM and timeout for both seqr and hail aggregation functions.


dancoates (approved):
Just out of curiosity, how much further can these go before we hit hard limits?

milo-hyben:
60 mins as timeout, with RAM not sure, but we are getting close to what we need, only last week or so there was a lot of large batches. Atm it is running every 2hrs and processing last 4 hours of batches, we can change to process 2hrs of batches every hour, which would be less data if needed. For time being I will monitor alerts and adjust RAM/timeouts as needed.

PR #268, opened 2024-09-24
New AR-config-viewer group

milo-hyben:
This new group is attached to cpg-config bucket.
All the analysis groups will be added as members, instead of directly attached to cpg-config bucket.
This will prevents 250 hard GCP limit on attached groups per resource we are currently experiencing.
Deployment will be done in steps.

1. Creating new group and attach all the existing analysis groups to it. (this PR)
2. Test, manually remove fewgenoms ar-groups from the cpg-config bucket and and submit a batch via AR, check for any permissions issues
3. If test is successful, then remove code on line 2603 as per [TODO comment](https://github.com/populationgenomics/cpg-infrastructure/blob/7985ece94fca4c63e2b2306c8aaee7e68e1f4ce1/cpg_infra/driver.py#L2603)

### DIFF
--- a/cpg_infra/driver.py
+++ b/cpg_infra/driver.py
@@ -2545,11 +2545,7 @@ class CPGDatasetCloudInfrastructure:
 
     @cached_property
     def config_viewer_group(self):
-        return self.group_provider.create_group(
-            self.common_gcp_infra,
-            cache_members=False,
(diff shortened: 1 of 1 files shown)


nevoodoo (approved):
LGTM!

PR #270, opened 2024-09-29
SET-170: move some functions to utility

ckrishk:
Moving function to utility file

### DIFF
--- a/cpg_infra/driver.py
+++ b/cpg_infra/driver.py
@@ -3,20 +3,18 @@
 CPG Dataset infrastructure
 """
 
+import graphlib
 import os.path
 import re
 from collections import defaultdict
 from functools import cached_property
 from typing import Any, Callable, Iterable, Iterator, NamedTuple, Type
 
-import graphlib
+import cpg_utils.config
 import pulumi
 import pulumi_gcp as gcp
 import toml
-import xxhash
-from toml_sort import TomlSort
 
-import cpg_utils.config
 from cpg_infra.abstraction.azure import AzureInfra
 from cpg_infra.abstraction.base import (
     BucketMembership,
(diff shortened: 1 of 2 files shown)


jmarshall:
Test action fails with:

```
ruff-format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

3 files reformatted, 40 files left unchanged
[…]
mypy.....................................................................Failed
[…]
cpg_infra/driver.py:71: error: Only concrete class can be given where
"type[CloudInfraBase]" is expected  [type-abstract]
        c.name(): c
```

I was able to reproduce this locally by setting up pre-commit. Here is what I think is going on here:

The original code on **main** is 

```python
    c.name(): c  # type: ignore
    for c in CloudInfraBase.__subclasses__()
```

Apparently the types here are too mysterious for mypy, hence the `type: ignore` annotation.

Your branch has reformatted that onto one line:

```python
    c.name(): c for c in CloudInfraBase.__subclasses__()  # type: ignore
```

However for whatever reason ruff-format doesn't like that and has reformatted it back again:

```python
    c.name(): c
    for c in CloudInfraBase.__subclasses__()  # type: ignore
```

Handily the later pre-commit hook apparently sees that uncommitted code rather than what's in your PR! And now the annotation is on the wrong line so is ineffective.

jmarshall:
I feel like you've got a case of duelling linters and autoformatters here. For example, your PR changes all the `'…'` strings to `"…"` — and the pre-commit ruff-format hook is changing them all back again. So both these strings and the formatting of the `c.name(): c for c in …` line were perhaps altered by something you've got set up locally, and it probably would benefit from being configured to agree more with the linters the code is already formatted for…

ckrishk:
Awesome thanks!! will give the linter change a try. I do remember the file
was saved without IDE changes. I will try removing it to see if that works.

PR #277, opened 2024-10-14
Increase RAM for billing aggregation functions.

milo-hyben:
This PR includes change to timeout for all billing aggregation to 1H.
Seqr & Hail functions RAM increased to 5GB.


violetbrina (commented):
Do you need to set memory for gcp separately to something lower? Or is the default memory value enough?

milo-hyben:
> Do you need to set memory for gcp separately to something lower? Or is the default memory value enough?

GCP seems to be fine with 1GB, from the logs it use up to 62% of allocated RAM in the last 2 weeks. 
I do not think lowering is needed.

violetbrina (approved):

PR #280, opened 2024-11-28
Adding images-tmp container repo

nevoodoo:
## Description

This adds a new tmp registry, just not sure if this is the way to do it so please offer any feedback on how to approach :)

### DIFF
--- a/cpg_infra/driver.py
+++ b/cpg_infra/driver.py
@@ -2524,15 +2524,15 @@ class CPGDatasetCloudInfrastructure:
             membership=ContainerRegistryMembership.WRITER,
         )
         self.infra.add_member_to_container_registry(
-            'test-full-reader-in-dev-container-registry',
+            'images-reader-in-tmp-container-registry',
             registry=tmp_container_registry,
-            member=self.test_full_group,
+            member=self.images_reader_group,
             membership=ContainerRegistryMembership.READER,
         )
         self.infra.add_member_to_container_registry(
-            'analysis-writer-in-dev-container-registry',
+            'images-writer-in-tmp-container-registry',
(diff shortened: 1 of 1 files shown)


MattWellie (approved):
I have no idea either :D Looks legit though, this should add the existing groups to the new registry...???

nevoodoo:
> I have no idea either :D Looks legit though, this should add the existing groups to the new registry...???

That is my assumption, I just copied whatever we were doing for images-dev :/

cpg_infra/driver.py
```diff
@@ -2522,6 +2523,18 @@ def setup_dataset_container_registry(self):
             member=self.analysis_group,
             membership=ContainerRegistryMembership.WRITER,
         )
+        self.infra.add_member_to_container_registry(
+            'test-full-reader-in-dev-container-registry',
```
milo-hyben, line 2527: This should be 'test-full-reader-in-tmp-container-registry', otherwise pulumi won't create as new I think.

cpg_infra/driver.py
```diff
@@ -2522,6 +2523,18 @@ def setup_dataset_container_registry(self):
...
+            registry=tmp_container_registry,
+            member=self.test_full_group,
+            membership=ContainerRegistryMembership.READER,
+        )
+        self.infra.add_member_to_container_registry(
+            'analysis-writer-in-dev-container-registry',
```
milo-hyben, line 2533: The same here should be analysis-writer-in-tmp-container-registry'

milo-hyben (changes requested):
I think this the right way, just you need to name member pulumi resource differently.

milo-hyben (approved):

PR #286, opened 2024-12-03
Add specific group for metadata contributors

dancoates:
This is only used to allow commenting within metamist without having metadata write access or being in the analysis group


milo-hyben (approved):
Hey @dancoates, this looks good to me, very consistent with metadata-access style.

PR #287, opened 2024-12-10
Temporarily add workflow.default_jar_spec_revision to infrastructure.toml

jmarshall:
Until upstream Hail releases proper fixes for SET-261 and especially SET-247, we need to override the QoB JAR file always.

### DIFF
--- a/cpg_infra/driver.py
+++ b/cpg_infra/driver.py
@@ -631,6 +631,11 @@ class CPGInfrastructure:
             # nest in .infrastructure
             d = {'infrastructure': dict(zip(keys, values))}
 
+            # JohnM 10 Dec 2024: Temporarily override QoB JAR to avoid ballooning hail costs
+            # cc7d0f is 0.2.132 with retry-all-400-errors (SET-261) and no 20million fixes (SET-247)
+            v132_retry_no20million = 'cc7d0fa48bd88b10b8c1c7315d9604ee855693e7'
+            d['workflow'] = {'default_jar_spec_revision': v132_retry_no20million}
+
             return dict_to_toml(d)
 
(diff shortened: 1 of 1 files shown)


violetbrina (approved):

jmarshall:
This is admittedly fairly hackily done, but we also hope to be able to remove it again within weeks.

violetbrina:
I don't remember if a merge to main in `cpg-infrastructure` triggers a redeploy or not. If not, you'll need to manually trigger it in the workflows of `cpg-infrastructure-private`.

cpg_infra/driver.py
```diff
@@ -631,6 +631,11 @@ def _build_config(values: list) -> str:
             # nest in .infrastructure
             d = {'infrastructure': dict(zip(keys, values))}
 
+            # JohnM 10 Dec 2024: Temporarily override QoB JAR to avoid ballooning hail costs
+            # cc7d0f is 0.2.132 with retry-all-400-errors (SET-261) and no 20million fixes (SET-247)
+            v132_retry_no20million = 'cc7d0fa48bd88b10b8c1c7315d9604ee855693e7'
```
vivbak, line 636: I really like how I understand what this variable is hahah

vivbak (approved):

cpg_infra/driver.py
jmarshall: The variable is only there because it wanted to reformat the long line when the constant was written out in the line below… 😂

violetbrina: Great variable naming for sure

jmarshall:
Thanks, I did indeed need to manually trigger the cpg-infrastructure-private workflow.

PR #291, opened 2025-01-16
Created backup script for all the billing BQ views.

milo-hyben:
This is PR is just backup of all existing BQ materialised views. Reason for this not as a pulumi code directly is due to pulumi limitations regarding update of existing BQ views. Views always have to be deleted and then recreated, which is very painful.
There was a suggestions to use this SQL scripts and running them as a separate tasks via pulumi if hash of the files are different. Something we can address at some stage.
At the moment those files are applied manually.


violetbrina (approved):
I thought this had already merged! Happy to approve, glad to have a record of our queries.

PR #292, opened 2025-01-20
Update QoB JAR override to include 20million fix and improved 400 logging

jmarshall:
Use a QoB JAR built today via a Hail CI deploy batch, which includes a fix for the "20 million string length" issue cherry-picked from an as-yet unreleased upstream fix (hail-is/hail#14129) and added logging to help identify the intermittent 400 Bad Request errors encountered locally (populationgenomics/hail#358).

This is a followup to #287 and #288. With a `default_jar_spec_revision` that includes fixes for all currently known QoB JAR problems, this will mean that individual pipelines don't need to specify their own `jar_spec_revision`.


milo-hyben (approved):
Looking good to me!

PR #297, opened 2025-03-17
Add images-dev-reader and images-dev-writer groups.

milo-hyben:
This is PR implements suggestion to control access to Artifact registry based on access level, e.g. dev/test account can't read from production Artifact Registry and production accounts can't read from test/dev AR.

### DIFF
--- a/cpg_infra/driver.py
+++ b/cpg_infra/driver.py
@@ -1125,8 +1125,16 @@ class CPGDatasetCloudInfrastructure:
         return self.create_group('images-reader')
 
     @cached_property
+    def images_dev_reader_group(self):
+        return self.create_group('images-dev-reader')
+
+    @cached_property
     def images_writer_group(self):
         return self.create_group('images-writer')
+
+    @cached_property
+    def images_dev_writer_group(self):
+        return self.create_group('images-dev-writer')
 
     @cached_property
     def analysis_group(self):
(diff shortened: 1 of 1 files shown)


vivbak:
Just a note that we need to time the roll out of this (so it's communicated before merge!)

vivbak:
We'll need to figure out the order in which we remove images-tmp (i.e. here first or in GCP first)

cpg_infra/driver.py
```diff
@@ -2504,13 +2512,27 @@ def setup_container_read_write_permissions(self):
             self.infra.get_pulumi_name('full-in-images-writer-group-member'),
             self.full_group,
         )
+        self.images_dev_writer_group.add_member(
+            self.infra.get_pulumi_name('dev-images-writer-group-member'),
```
jmarshall, line 2516: Is this right? This pulumi is clear as mud to me, but other similar things seem to follow the pattern `'dev-in-images-writer-group-member'`.

milo-hyben: Good point @jmarshall, by reviewing existing groups I can see there are no actual 'dev' groups, there are only 'test-full' and 'test'. I had adjusted the get_pulumi_name calls.

jmarshall (approved):
Thanks @milo-hyben, I'll approve this on a suck-it-and-see approach.

One more note here that it's approved but we won't merge it until we've sorted out the deployment plan.

PR #298, opened 2025-04-16
Add images_location_format entry recording production images path

jmarshall:
Made available in the config as `infrastructure.images_location_format`, similarly to `git_credentials_secret_name`.


jmarshall:
This is no longer necessary.

PR #299, opened 2025-05-12
Increase CPU to 1, give storage_visualization more RAM.

milo-hyben:
This PR implements changes cpu from 0.5 to 1, give storage_visualization around 7GB or RAM so it can complete seqr bucket calculations.

### DIFF
--- a/storage_visualization/disk_usage.py
+++ b/storage_visualization/disk_usage.py
@@ -110,7 +110,9 @@ def count_stats_for_bucket(
                 count += 1
                 if count % 10**6 == 0:
                     s = naturalsize(sys.getsizeof(aggregate_stats_container))
-                    logging.info(f'{count // 10**6} M blobs... aggregate dict is using {s}')
+                    logging.info(
+                        f'{count // 10**6} M blobs... aggregate dict is using {s}'
+                    )
                 folder = f'/{aggregate_level(blob.name)}'
                 while True:
                     path = f'gs://{bucket_name}{folder}'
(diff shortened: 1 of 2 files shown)


jmarshall:
I don't believe that's how this API works: because `list_buckets()` returns a `google.api_core.page_iterator.Iterator`, the iterator will handle the paging internally for you. At least, that was my experience using similar APIs from google-cloud-artifact-registry.

We should test whether this thing is already paging, and what page size it is using. It's processing millions of blobs (in the other buckets; I don't know what the count for the `-tmp` bucket is), so `page_size=100` may slow it down substantially.

milo-hyben:
> I don't believe that's how this API works: because `list_buckets()` returns a `google.api_core.page_iterator.Iterator`, the iterator will handle the paging internally for you. At least, that was my experience using similar APIs from google-cloud-artifact-registry.
> 
> We should test whether this thing is already paging, and what page size it is using. It's processing millions of blobs (in the other buckets; I don't know what the count for the `-tmp` bucket is), so `page_size=100` may slow it down substantially.


I guess I should try to run locally and check memory. Paging was recommended. 3.5GB is a lot for just iterating over filenames and small amount of metadata. Even for large amount of files. Let me do some testing locally.

jmarshall:
I think it will be useful to decrease the `count % 10**6` to `count % 50000`or so, so that it prints out the `aggregate_stats_container` size more often so we can see it exploding, and to see if there is some previously unobserved kind of file tree that needs to be added to `aggregate_level()`.

milo-hyben:
I had tried a little test with thousand-genomes and paging vs not paging does not make any difference to peak memory.
I guess all the memory would be consumed by 'aggregate_stats_container'.
I can 2 fields 'STANDARD_bytes' and 'size' having the same value, wonder if need both. e.g from thousand-genomes:

aggregate_stats_container contains:
'gs://cpg-thousand-genomes-main/cuking/v0': defaultdict(<class 'int'>, {'size': **79973152**, 'STANDARD_bytes': **79973152**, 'num_blobs': 3202, 'monthly_storage_cost': 0.0017130584418774512}),

milo-hyben:
I had done more tests locally and reported RAM in the logs is not correct, it is using getsizeof(), which is basically 'shallow' size of the object. I had used tracemalloc locally to get better size and here are differences:

So when at reports 83.9MB, it is already using 1.4.GB, from my local logging shows:
```
INFO:root:8 M blobs... aggregate dict is using 83.9 MB
Current memory usage: 1408.81 MB
Peak memory usage: 1413.96 MB
```

From the failed  cpg-seqr-main-tmp till the end it needs extra around 0.8GB, so we really need 4.5GB of RAM.
From Hail documentation I can see for highmem, it is uses 7GB per 1 core.
I had changed this PR to increase CPU from 0.5 to 1.0. Changing the title of PR as well.

jmarshall (approved):
The simplest change for this :smile:

Don't forget to delete all the irrelevant commit messages when squashing…

PR #301, opened 2025-05-21
Update QoB JAR override to be 0.2.134-based and include 400logging

jmarshall:
Update the QoB JAR override again, primarily to make it correspond to the deployed 0.2.134 hail release. This is a followup to #287, #288, and #292. We are nearly at a point where we won't need or want to override the QoB JAR by default anymore.

Use a QoB JAR built via a Hail CI deploy batch. This includes release 0.2.134 (which fixes the "20 million string length" issue) and retains our added logging to help identify the intermittent 400 Bad Request errors previously encountered locally (SET-261).

The selected commit, 2bb197d63afd58f9452af306e0eb14d6924d0019, is 0.2.134.cpg1 as tagged for CPG image building purposes (cf populationgenomics/images#200 and populationgenomics/images#201).

### DIFF
--- a/cpg_infra/driver.py
+++ b/cpg_infra/driver.py
@@ -632,10 +632,9 @@ class CPGInfrastructure:
             d = {'infrastructure': dict(zip(keys, values))}
 
             # JohnM 10 Dec 2024: Temporarily override QoB JAR to avoid ballooning hail costs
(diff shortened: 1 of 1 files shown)


vivbak (approved):

jmarshall:
We'll need to run the cpg-infrastructure-private workflow to actually deploy this, which I'll do when I've sorted out one other thing that that will want to deploy at the same time.

PR #302, opened 2025-05-27
Increasing BQ Insert Chunks, gzipping data before load, upgrading pan…

milo-hyben:
…das to_gbq.

### DIFF
--- a/cpg_infra/billing_aggregator/aggregate/utils.py
+++ b/cpg_infra/billing_aggregator/aggregate/utils.py
@@ -4,6 +4,7 @@ Class of helper functions for billing aggregate functions
 """
 
 import asyncio
+import gzip
 import json
 import logging
 import math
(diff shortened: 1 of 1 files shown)


nevoodoo (approved):
Looks good! Would switching to [Polars](https://pola.rs/) be beneficial in any way to speed operations? I haven't done a deep dive of the code so I don't know how actively `pandas` is being used but worth considering a switch

dancoates:
> Looks good! Would switching to [Polars](https://pola.rs/) be beneficial in any way to speed up operations? I haven't done a deep dive of the code so I don't know how actively `pandas` is being used but worth considering a switch

That's a good callout, bigquery can return results as arrow format, which polars can read very efficiently eg.

```py
dataframe = pl.from_arrow(result.to_arrow())
```

It can then write back to bigquery as described here:

https://docs.pola.rs/user-guide/io/bigquery/#write

Could be worth exploring if we're still hitting resource constraints.

PR #308, opened 2025-06-09
Decompress 'sequencing_groups_gzip' job attribute if present

jmarshall:
Aggregator-side implementation for SET-565. Handle large `sequencing_groups` job attributes by compressing them for transport via the Hail database as a `sequencing_groups_gzip` attribute and decompressing them here.


milo-hyben (approved):
Looking good to me!

PR #309, opened 2025-06-16
Trim depth of recursion for VDS in storage-vis

MattWellie:
Our storage plots have gone haywire - from a 5MB HTML file a year ago to 500+MB now, these can't be rendered. 

My suspicion is that the Hail Object detection (highly nested directories) only applies to HT/MT, and was not aware of VDSs. A VDS is a direcotry containing two MTs, but stored as `reference_data` and `variant_data`, so they wouldn't independently be recognised by the `.mt` extension detection. 

This might be the one secret ingredient to make these reports usable again, or there may be some other underlying cause for the bloat, but in either scenario this feels like a good change to make.

### DIFF
--- a/storage_visualization/disk_usage.py
+++ b/storage_visualization/disk_usage.py
@@ -56,8 +56,8 @@ def aggregate_level(name: str) -> str:
         return name[: mt_index + 3]
     if (vds_index := name.find('.vds/')) != -1:
         return name[: vds_index + 4]
-    if (batch_tmp_index := name.rfind('batch-tmp')) != -1:
-        return name[: batch_tmp_index + 9]
+    if (batch_tmp_index := name.rfind('/batch-tmp')) != -1:
+        return name[: batch_tmp_index + 10]
     if (slash_index := name.rfind('/')) != -1:
         return name[:slash_index]
     return ''  # Root level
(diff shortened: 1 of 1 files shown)


storage_visualization/disk_usage.py
```diff
@@ -54,6 +54,8 @@ def aggregate_level(name: str) -> str:
         return name[: ht_index + 3]
     if (mt_index := name.find('.mt/')) != -1:
         return name[: mt_index + 3]
+    if (mt_index := name.find('.vds/')) != -1:
+        return name[: mt_index + 4]
```
jmarshall, line 58:
```suggestion
    if (vds_index := name.find('.vds/')) != -1:
        return name[: vds_index + 4]
```

jmarshall (approved):
Dumb comment re variable name, otherwise LGTM.

storage_visualization/disk_usage.py
MattWellie: spotted this as soon as I published the PR 🤦

storage_visualization/disk_usage.py
```diff
@@ -56,6 +56,8 @@ def aggregate_level(name: str) -> str:
         return name[: mt_index + 3]
     if (vds_index := name.find('.vds/')) != -1:
         return name[: vds_index + 4]
+    if (batch_tmp_index := name.rfind('batch-tmp/')) != -1:
```
jmarshall, line 59: All the manky stuff that's recently appeared in cpg-*-test-tmp actually matches `/batch-tmpNNNNNN/` where the NNNNNN appears to be six hex digits. So alas this won't catch it.

(And we should figure out what's not following the pattern!)

PR #315, opened 2025-08-21
Update to Billing Materialised Views to reflect changes to Seq group and Cohorts

milo-hyben:
This PR is just a backup of Big Query materialised views. Views have been already updated.


jmarshall (approved):
Looks plausible to me, though I note that [`JSON_EXTRACT_STRING_ARRAY ` is apparently deprecated](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_extract_string_array) and one is supposed to use `JSON_VALUE_ARRAY` instead — which appears to have the same behaviour, so it would just be a simple renaming.

PR #316, opened 2025-09-09
Replacing ARRAY_TO_STRING with REGEXP_REPLACE to include the old style sequencing_groups, where encoded inverted commas are present.

milo-hyben:
This PR is a simple fix to the old style sequencing_groups with encoded inverted commas are present.


violetbrina (commented):
Nice update, I know the files here are more for documentation/reference to the tables. 
It's unrelated to this PR but the tests seem to be failing? Not sure why 🤷‍♀️

My only hesitation would be to merge a PR with failing tests into main, even though this really shouldn't be affecting that.

violetbrina:
> Nice update, I know the files here are more for documentation/reference to the tables. It's unrelated to this PR but the tests seem to be failing? Not sure why 🤷‍♀️
> 
> My only hesitation would be to merge a PR with failing tests into main, even though this really shouldn't be affecting that.

I guess if there aren't any tests, we should probably just remove or disable the test workflow.... kind of silly that it's here if it's not doing anything 🤷‍♀️

violetbrina (approved):
LGTM

PR #319, opened 2025-10-02
Add RD specific storage visualisation code and workflow

EddieLF:
For running the automated storage visualisation, but scoped to just datasets in the `'rare-disease'` dependency list. 

- A new entrypoint was added which filters the datasets to just the RD ones, and uses the `'rare-disease'` dataset to submit to analysis runner.
- The Github workflow was duplicated from the existing one, only changing the entrypoint.
- A new config `rd_slack.toml` is added to set the Slack channel name, and the GCP secret project / name where the Slack API token is stored. 

We will need to create the Slack channel, Slack bot, and GCP secret for the bot to work.

### DIFF
--- a/.github/workflows/rd_storage_visualization.yaml
+++ b/.github/workflows/rd_storage_visualization.yaml
@@ -1,9 +1,9 @@
-name: Storage visualization
+name: RD storage visualization
 
 on:
-  # Run every Sunday morning.
+  # Run every Monday morning.
   schedule:
-    - cron: "0 0 * * 0"
+    - cron: "0 4 * * 1"
   # Run on demand.
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


.github/workflows/rd_storage_visualization.yaml
```diff
@@ -0,0 +1,49 @@
+name: Storage visualization
+
+on:
+  # Run every Sunday morning.
+  schedule:
+    - cron: "0 0 * * 0"
```
cassimons, line 6: can we make it like 4am on Monday. I  always end up looking at these on Sunday night when they come through and I really do not want to.

MattWellie:
I think this looks good, but there's a wild process here (in the original too):
- the workflow clones the whole of cpg-infrastructure-private
- it builds a new dockerfile containing that latest version of the config
- it runs the workflow inside that container, which has the config baked in, then reads and uses the config

The result here is that we make a new docker image every few days 🤮 https://console.cloud.google.com/artifacts/docker/cpg-common/australia-southeast1/images/storage-visualization 

The analysis-runner wrapper already has functionality to send config content as part of the API call, which feels like a much more intuitive way to accomplish this

PR #355, opened 2026-03-18
Add CPGInfrastructureGroup and creation of ad-hoc google groups

jmarshall:
Add an `adhoc_groups` field to `CPGInfrastructureConfig` that can be populated to have the driver create miscellaneous google groups that are not associated with datasets.


milo-hyben (approved):
This is looking good to me!

PR #357, opened 2026-04-16
Fix dependent PAM entitlements

nevoodoo:
This allows dependent datasets to have entitlements created automatically without requiring `allow_notebook_tmp_main_read` on every dataset in the dependency chain. Previous attempt tried to this directly from the requesting workflow but looks like we may need changes in the infra code after all.


milo-hyben (approved):
Looks fine to me!

PR #359, opened 2026-05-22
Enable multiple parallel metamist services (e.g. production and development) in config

violetbrina:
Currently we can only support the permissions and service account access for a single deployment of metamist.

While it is normative that we only have one, during the process of the metamist postgres migration it became clear we needed to support two parallel deployments. 

These permissions that will be deployed by pulumi will allow our current hail accounts and users with their appropriate permissions to access the metamist-development service in the metamist-dev project.

Removing the extra service in config in future should revoke these permissions.

Why do this change at all with Dan's modification for the legacy account? Truly the only critical change was having the `infra.add_cloudrun_invoker` for both the new and legacy accounts. But that call needs not only the machine account but the servcie name and the project name, requiring a refactor.

The justification for the change to a list means that in future we only need to modify the cpg-infrastructure-private config file to remove and add any number of metamist services. The driver will not need to be modified.

Corresponding PR in cpg-infrastructure-private: https://github.com/populationgenomics/cpg-infrastructure-private/pull/695

### DIFF
--- a/cpg_infra/billing_aggregator/monthly_aggregate/main.py
+++ b/cpg_infra/billing_aggregator/monthly_aggregate/main.py
@@ -128,7 +128,7 @@ async def process_and_upload_monthly_billing_report(
 
     data['cost'].fillna(0)
     data['key'] = data.topic + '-' + data.month + '-' + data.cost_category
-    values: list = data.to_numpy().tolist()
+    values: list = list(data.to_numpy().tolist())
     updated = append_values_to_google_sheet(OUTPUT_GOOGLE_SHEET, values, invoice_month)
 
     return f'{updated} cells appended for invoice month {invoice_month}', 200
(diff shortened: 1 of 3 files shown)


cpg_infra/billing_aggregator/monthly_aggregate/main.py
```diff
@@ -128,7 +128,7 @@ async def process_and_upload_monthly_billing_report(
 
     data['cost'].fillna(0)
     data['key'] = data.topic + '-' + data.month + '-' + data.cost_category
-    values: list = data.to_numpy().tolist()
+    values: list = list(data.to_numpy().tolist())
```
violetbrina, line 131: A mypy issue that arose with typing. Resolved with explicit casting.

rameshka (approved):
LGTM @violetbrina !!. Thanks for the very informative PR description.

PR #368, opened 2026-06-12
Add metamist infra changes back

rameshka:
This PR revert the changes made in https://github.com/populationgenomics/cpg-infrastructure/pull/367
Pulumi preview - https://github.com/populationgenomics/cpg-infrastructure-private/actions/runs/28628754057/job/84900932794

**Additional info**

Currently we can only support the permissions and service account access for a single deployment of metamist.

While it is normative that we only have one, during the process of the metamist postgres migration it became clear we needed to support two parallel deployments.

These permissions that will be deployed by pulumi will allow our current hail accounts and users with their appropriate permissions to access the metamist-development service in the metamist-dev project.

Removing the old service in config in future should revoke these permissions.

Why do this change at all with Dan's modification for the legacy account? Truly the only critical change was having the infra.add_cloudrun_invoker for both the new and legacy accounts. But that call needs not only the machine account but the servcie name and the project name, requiring a refactor.

Corresponding PR in the cpg-infra-private https://github.com/populationgenomics/cpg-infrastructure-private/pull/701

### DIFF
--- a/cpg_infra/config/config.py
+++ b/cpg_infra/config/config.py
@@ -160,10 +160,24 @@ class CPGInfrastructureConfig(DeserializableDataclass):
     class Metamist(DeserializableDataclass):
         @dataclasses.dataclass(frozen=True)
         class GCP(DeserializableDataclass):
+            @dataclasses.dataclass(frozen=True)
+            class GCPMetamistDeploys(DeserializableDataclass):
+                project: str
+                service_name: str
+                machine_account: str
+
+            # Weird structure explained:
+            # The project service_name and machine account are left as-is
+            # so the metamist etl infrastructure code defined in the metamist
+            # repo still work. Any additional deploys go into the dev list
+
+            # This way all of the infra.metamist.gcp.project references in the
+            # etl plugin will continue to work
             project: str
             service_name: str
             machine_account: str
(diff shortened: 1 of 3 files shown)


cpg_infra/config/config.py
```diff
@@ -160,11 +160,26 @@ class GCP(DeserializableDataclass):
...
+                service_name: str
+                machine_account: str
+
+            # Weird structure explained:
+            # The project service_name and machine account are left as-is
+            # so the metamist etl infrastructure code defined in the metamist
+            # repo still work. Any additional deploys go into the dev list
+
+            # This way all of the infra.metamist.gcp.project references in the
+            # etl plugin will continue to work
```
rameshka, line 175: The [new config](https://github.com/populationgenomics/cpg-infrastructure-private/pull/701/changes) defined here maps to this `GCP` class now.

As per the above comment, for the ETL plugin to work when decommissioning the `legacy_machine_account`, 

- We will need to map the `prod` `GCPMetamistDeploys` to the level where `legacy_machine_account` is in the config file
- Remove the `prod` config

milo-hyben (approved):
Thank you Piyumi! Looking much better now.

PR #372, opened 2026-06-26
[SET-1051] Highlight projects with unlinked billing in the cost report

violetbrina:
## What & why

The daily GCP cost report (`slack_bot_cost_report`) now highlights projects whose **billing account has been unlinked/disabled**, so they aren't forgotten and left at risk of data deletion (Google can delete data from a project that has no billing account attached). Previously the only signal was the one-shot message `gcp_cost_control` posts when it auto-disables billing on a budget breach — easy to miss, and the risk persists until billing is re-linked.

Resolves **SET-1051**. Also folds in **SET-588** (hail row appearing twice when over budget).

## Changes

**Unlinked-billing highlight (SET-1051)**
- New `billing_status.py`:
  - `get_unlinked_project_ids(...)` — checks each candidate via `google-cloud-billing`'s `CloudBillingClient.get_project_billing_info`. Three-state: a project is flagged **only** on a definitive `billing_enabled == False`; any API error (e.g. a 403 on external projects the service account can't see, or a transient failure) is skipped and logged at INFO, so a hiccup never produces a false "unlinked" alert.
  - `apply_unlinked_to_summary(...)` — rewrites unlinked projects' rows.
- Candidate set = **union** of projects with a budget and projects in the cost report (excluding the `<none>` bucket). The budget-map side catches projects that have dropped out of the billing export because they have no recent cost.
- Unlinked projects sort to the **top of the flagged list**, rendered **bold**, with `💵⛓️‍💥 Billing unlinked` in the Previous Day column and their normal month value preserved (or `No monthly cost`).
- `google-cloud-billing` added to `requirements.txt`. No new IAM — runs as the existing `gcp-cost-control` service account, which can read billing status for CPG-owned projects.
- `gcp_cost_control` is **untouched** — its `is_billing_enabled` is intentionally not shared, to avoid any risk to that production function.
- Added a `README.md` documenting a dev deploy/test loop.

**Bold + hail dedupe (SET-588)**
- All unlinked projects are bolded (link, indicator, and month column) to match the emphasis of the other flagged rows.
- The hail project is pinned to the flagged section only when it isn't already flagged. When hail is over budget it's already present (and bolded by `format_billing_row`), so the previous unconditional append listed it twice.

## Testing

Deployed to a dev Cloud Function (`gcp-cost-reporting-dev`, same service account, posting to a dev Slack channel) and triggered it. Verified the rendering end-to-end — the successful message:

> We are 87% through the month. Costs exceeding the monthly budget or daily limit (6.7%) are flagged below:

| Projects | Previous Day (%) \| Month (%) |
| --- | --- |
| **ravenscroft-rpl** | **💵⛓️‍💥 Billing unlinked** \| **$101 (67%)** |
| **udn-aus-shared** | **💵⛓️‍💥 Billing unlinked** \| **No monthly cost** |
| aasgard-dev | $33.26 (1%) \| $3750 (94%) |
| hail-295901 | $607 (2%) \| $6032 (17%) |

`ravenscroft-rpl` (has cost) and `udn-aus-shared` (budget-only, no cost — caught via the budget-map side of the union) both surface at the top in bold; `hail-295901` appears once.

> Note: this function is deployed to production by Pulumi (`cpg_infra/billing_aggregator/driver.py`), so it ships on the next infra apply, not on merge alone.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


milo-hyben (approved):
Looking good @violetbrina !

PR #376, opened 2026-07-14
Switch to use pydantic for config validation

dancoates:
Previously config was deserialised using a custom dataclass, this was performing a function that pydantic does really well and isn't really necessary to have our own custom code for. Using pydantic gives us better handling of different types, and a more strict validation of config.

This should produce no infra changes, we will need to verify this by running a `preview` from [cpg-infrastructure-private](https://github.com/populationgenomics/cpg-infrastructure-private/)


cpg_infra/config/base.py
```diff
@@ -0,0 +1,24 @@
...
+from pydantic import BaseModel, ConfigDict
+
+
+class ConfigModel(BaseModel):
+    """Base for all CPG infra config models: validates on construction, immutable."""
+
+    model_config = ConfigDict(frozen=True, extra='forbid')
+
+    @classmethod
+    def from_dict(cls: type[Self], d: dict[str, Any]) -> Self:
```
dancoates, line 22: This alias is left here as `from_dict` is used in metamist, so keeping this for compatibility means that we don't need to go and update metamist ETL which is a pain https://github.com/populationgenomics/metamist/blob/main/metamist_infrastructure/__main__.py#L76

violetbrina (commented):
I've had a look through, almost all of the changes are quite trivial syntax-wise so it's pretty easy to see the 1-for-1 changes to using the Pydantic BaseModel (and our ConfigModel built on that) throughout. 

My main question here is on the preview. The change in the retry policy could very well(?) have been intentional to remove the no retry. But these config values here being removed do concern me a bit. Is there a change in config that explains this?

```
- retryPolicy: "RETRY_POLICY_DO_NOT_RETRY" (probably a good one to remove)
```

Just a bit concerned about these being removed, can we confirm that they aren't needed/this is an intentional change?
```
- BIGQUERY_LOG_TABLE       : ...
- BIGQUERY_TABLE           : ...
- CONFIGURATION_SECRET     : ...
- NOTIFICATION_PUBSUB_TOPIC: ...
- PUBSUB_TOPIC             : ...
- SM_ENVIRONMENT           : ...
```

I'm aware the preview produce the same result to main, but that just may mean a config change in infra-private isn't reflected in our infra yet. If this is meant to be applied happy to approve. Just curious...

rameshka (approved):
Hey, @dancoates, proposed changes look good.

violetbrina (approved):
Great. Thanks for that sanity check. LGTM.

PR #378, opened 2026-07-28
[SET-1228] Sync metamist project display name & description from cpg-infra

violetbrina:
## Summary

Implements [SET-1228](https://cpg-populationanalysis.atlassian.net/browse/SET-1228): let the RD team set a stylised **display name** and optional **description** on metamist projects from cpg-infrastructure dataset config.

Because metamist projects have no first-class columns for these (deferred until the Postgres migration, per the ticket's No-Gos), the values are stored in the project's `meta` JSON under keys `display_name` and `description`.

## Changes

- **`config.py`** — two optional fields on `CPGDatasetConfig`: `display_name`, `description` (only apply when `enable_metamist_project` is true).
- **`abstraction/metamist.py`** — `MetamistProject`/`MetamistProjectProvider` now take a generic `meta` dict:
  - `create` writes meta via a follow-up `update_project` call (create_project can't carry meta), for both newly-created **and** pre-existing projects.
  - new `update` method → meta-only changes are in-place updates, never project replaces (`diff` routes them there).
  - **Declarative clearing**: keys dropped from config are sent as explicit JSON `null`, because the server merge-patches meta (an empty `{}` is a no-op).
- **`driver.py`** — builds the meta dict from config; the test project gets a `(test)`-suffixed display name.
- **`test/test_config_validation.py`** — coverage for the new fields (default `None`, parse when provided).

## Design record

ADR (kept in Confluence, not the repo): https://cpg-populationanalysis.atlassian.net/wiki/spaces/ST/pages/1384218630

## Verification

- `test/test_config_validation.py` — 9 passed (incl. new `test_metamist_project_labels`).
- All three edited modules compile; declarative-null merge + `(test)` suffix logic verified against sample inputs.
- ⚠️ To confirm on review: the credential cpg-infra uses must have `project_admin` + project-creator rights for `update_project` (it already calls `create_project`/`update_project_members`, so almost certainly yes).

## Known limitations

- **No server-drift reconciliation.** `read` doesn't re-fetch server-side meta, so `diff` compares only against last-recorded config state. If a project's `display_name`/`description` is edited directly in metamist (outside cpg-infra), the drift is **not** detected or corrected. This is by design: these values are expected to be changed in config only — cpg-infra config is the source of truth.

[SET-1228]: https://cpg-populationanalysis.atlassian.net/browse/SET-1228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cpg_infra/abstraction/metamist.py
+++ b/cpg_infra/abstraction/metamist.py
@@ -50,7 +50,7 @@ class MetamistProjectProvider(pulumi.dynamic.ResourceProvider):
         if not project_id:
             raise RuntimeError(f'Failed to create project {name}')
 
-        # create_project can't carry meta, so sync it with a follow-up call.
+        # create_project doesn't carry meta, so sync it with a follow-up call.
         # Applies to both newly-created and pre-existing projects.
         if meta:
             ProjectApi().update_project(name, {'meta': meta})
@@ -69,14 +69,19 @@ class MetamistProjectProvider(pulumi.dynamic.ResourceProvider):
         old_meta = _olds.get('meta') or {}
         new_meta = _news.get('meta') or {}
 
-        # Declarative sync: the server merge-patches meta, so a key dropped from
-        # config would otherwise linger. Send an explicit null for any key that
-        # was set before but is absent now, so it gets cleared.
-        merged = {k: None for k in old_meta}
-        merged.update(new_meta)
+        # The cpg-infra-private config is the source of truth for these meta keys:
+        # we sync config -> metamist, but never metamist -> config. So to change a
+        # display_name or description, edit the dataset config rather than editing
+        # the project meta directly in metamist.
+        #
+        # The server merge-patches meta, so a key dropped from the config would
+        # otherwise linger. Send an explicit null for those keys to clear them.
+        meta_update = dict(new_meta)
+        for k in old_meta.keys() - new_meta.keys():
+            meta_update[k] = None  # Tell update_project() to remove this entry
 
-        if merged:
-            ProjectApi().update_project(name, {'meta': merged})
+        if meta_update:
+            ProjectApi().update_project(name, {'meta': meta_update})
 
         return pulumi.dynamic.UpdateResult(
             outs={
--- a/cpg_infra/abstraction/metamist.py
+++ b/cpg_infra/abstraction/metamist.py
@@ -118,13 +118,13 @@ class MetamistProject(pulumi.dynamic.Resource):
 
     project_id: pulumi.Output[int]
     project_name: pulumi.Output[str]
-    meta: pulumi.Output[dict]
+    meta: pulumi.Output[dict[str, str]]
 
     def __init__(
         self,
         name: str,
         project_name: str,
-        meta: dict | None = None,
+        meta: dict[str, str] | None = None,
         opts: pulumi.ResourceOptions | None = None,
     ) -> None:
         args = {
--- a/cpg_infra/abstraction/metamist.py
+++ b/cpg_infra/abstraction/metamist.py
@@ -36,6 +36,7 @@ class MetamistProjectProvider(pulumi.dynamic.ResourceProvider):
 
     def create(self, props: dict[str, Any]) -> pulumi.dynamic.CreateResult:
         name = props['project_name']
+        meta = props.get('meta') or {}
 
         if project := get_project_by_name(name):
             project_id = project['id']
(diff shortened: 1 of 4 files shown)


cpg_infra/driver.py
```diff
@@ -1028,13 +1028,15 @@ def metamist_project(self):
...
             project_name=self.dataset,
+            meta=self.dataset_config.metamist_project_meta(),
         )
 
     @cached_property
     def metamist_test_project(self):
         return MetamistProject(
             f'metamist-project-{self.dataset}-test',
             project_name=self.dataset + '-test',
+            meta=self.dataset_config.metamist_project_meta(is_test=True),
```
jmarshall: The `-test` suffix is hardcoded in the lines above, so I'd just make the parameter `suffix` and write this as


```suggestion
            meta=self.dataset_config.metamist_project_meta(' (test)'),
```

cpg_infra/abstraction/metamist.py
```diff
@@ -49,11 +50,39 @@ def create(self, props: dict[str, Any]) -> pulumi.dynamic.CreateResult:
         if not project_id:
             raise RuntimeError(f'Failed to create project {name}')
 
+        # create_project can't carry meta, so sync it with a follow-up call.
```
jmarshall:
```suggestion
        # create_project doesn't carry meta, so sync it with a follow-up call.
```

It's a limitation of the existing implementation, not a law of the universe :smile:

cpg_infra/abstraction/metamist.py
```diff
@@ -49,11 +50,39 @@ def create(self, props: dict[str, Any]) -> pulumi.dynamic.CreateResult:
...
+    def update(self, _id: str, _olds, _news) -> pulumi.dynamic.UpdateResult:
+        name = _news['project_name']
+        old_meta = _olds.get('meta') or {}
+        new_meta = _news.get('meta') or {}
+
+        # Declarative sync: the server merge-patches meta, so a key dropped from
+        # config would otherwise linger. Send an explicit null for any key that
+        # was set before but is absent now, so it gets cleared.
+        merged = {k: None for k in old_meta}
+        merged.update(new_meta)
```
jmarshall: This makes `new_meta` look like an afterthought. I think I'd find something like the following clearer:

```
meta_update = new_meta
for k in old_meta.keys() - new_meta.keys():
    meta_update[k] = None  # Tell update_project() to remove this entry
```

test/test_config_validation.py
```diff
@@ -60,6 +60,81 @@ def test_dataset_config_example(self):
...
+            display_name='Rare Disease Cohort',
+            description='A cohort for rare disease research',
+        )
+        self.assertEqual(
+            {
+                'display_name': 'Rare Disease Cohort (test)',
+                'description': 'A cohort for rare disease research',
+            },
+            config.metamist_project_meta(is_test=True),
+        )
```
jmarshall: TBH I don't see the point of exhaustively testing `metamist_project_meta()` if we don't really have a way to test the whole flow. Especially if you apply my `metamist_project_meta(' (test)')` suggestion, the function is trivial, so testing all these possibilities is essentially equivalent to reading the code (and makes changing it harder).

cpg_infra/abstraction/metamist.py
```diff
@@ -49,11 +50,39 @@ def create(self, props: dict[str, Any]) -> pulumi.dynamic.CreateResult:
...
+                'meta': meta,
+            },
+        )
+
+    def update(self, _id: str, _olds, _news) -> pulumi.dynamic.UpdateResult:
+        name = _news['project_name']
+        old_meta = _olds.get('meta') or {}
+        new_meta = _news.get('meta') or {}
+
+        # Declarative sync: the server merge-patches meta, so a key dropped from
```
jmarshall: If you keep this comment, please clarify what “Declarative sync” is being used to mean.

jmarshall (commented):
I think it's basically fine. A few comments. Like any pulumi stuff, I guess we won't know if it works for sure until we try to run it end-to-end…

cpg_infra/driver.py
violetbrina: Done — `metamist_project_meta()` now takes `suffix: str = ''` and the call site passes `' (test)'`, matching the hardcoded `-test` above it. Simplifies the body to `meta['display_name'] = self.display_name + suffix`.

cpg_infra/abstraction/metamist.py
violetbrina: Fair 😄 — applied your wording.

cpg_infra/abstraction/metamist.py
violetbrina: Agreed, that reads much better — adopted, with one tweak:

```python
meta_update = dict(new_meta)
for k in old_meta.keys() - new_meta.keys():
    meta_update[k] = None  # Tell update_project() to remove this entry
```

`dict(new_meta)` rather than a bare alias, because `new_meta` is `_news['meta']` and is returned as-is in the `UpdateResult` outs below — aliasing it would leak the `None` sentinels into the recorded state.

test/test_config_validation.py
violetbrina: Agreed — removed all four `metamist_project_meta_*` tests (and the `_dataset_config` helper). With the `suffix` change the function is a couple of lines, so the tests just restated its body and would make it annoying to change.

Kept `test_metamist_project_labels`, which only covers that the two new config fields parse and default to `None`, in line with the other field-validation tests in that module.

cpg_infra/abstraction/metamist.py
violetbrina: Good catch — "declarative sync" was doing too much work in one phrase. What I meant is the sync direction: we have a cpg-infra-private config → metamist sync, but no metamist → infra config sync. The intended workflow is to modify the config; you're not meant to change this meta directly in metamist. Rewritten as:

```python
# The cpg-infra-private config is the source of truth for these meta keys:
# we sync config -> metamist, but never metamist -> config. So to change a
# display_name or description, edit the dataset config rather than editing
# the project meta directly in metamist.
#
# The server merge-patches meta, so a key dropped from the config would
# otherwise linger. Send an explicit null for those keys to clear them.
```

jmarshall (approved):

PR #379, opened 2026-08-05
Create external-repository-reader role

amyminiter:
### Summary
Added a new `external-repository-reader` role that grants members READ access to the main bucket only. Required to support access for a Terra Data Repository (TDR) dataset that requires service account access with READ permissions. 

### Changes
* `cpg_infra/config/config.py` — added 'external-repository-reader' to the GroupName literal so it's a valid key under members: in members.yaml.
* `cpg_infra/driver.py`:
    * New external_repository_reader_group cached property that creates the external-repository-reader group.
    * Group added to the groups list in setup_externally_specified_members so users listed under external-repository-reader: in members.yaml are added to it.
    * Direct READ binding on main_bucket (and only that bucket) added in setup_storage_main_bucket_permissions.

### External resources
* [ Slack conversation discussing the change ](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1785376464664549)
* [Proposal for Terra pilot](https://cpg-populationanalysis.atlassian.net/wiki/spaces/CP/pages/1349091390/Proposal+Terra+data+and+permissions+models)

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,8 +126,8 @@ Members have different roles within a dataset, those roles include:
 - upload:
   - Upload data into upload buckets as specified in the `datasets/$DATASET/upload.yaml` file.
 - external-repository-reader
-  - Read access to data in main. 
-  - Developed to support Terra Data Repository service account access to self-hosted production data in CPG buckets. 
+  - Read access to data in main.
+  - Developed to support Terra Data Repository service account access to self-hosted production data in CPG buckets.
 
 ## Setup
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -125,6 +125,9 @@ Members have different roles within a dataset, those roles include:
   - Access to view main-web bucket through web server
 - upload:
   - Upload data into upload buckets as specified in the `datasets/$DATASET/upload.yaml` file.
+- external-repository-reader
```
dancoates, line 128: Might be worth including a note here that this is distinct from data manager role as it _only_ grants read access to main and no additional access to any other buckets or services.

dancoates (approved):
This looks reasonable to me. Only other thing to note is that we'll need to be strict with enforcing that this new permission group is only used to grant external repository access and can't be used to grant elevated read access to human users.

MattWellie (approved):

MattWellie (approved):

PR #380, opened 2026-08-05
SET-1242 Add display_name, team_ownership & billing_groups to Dataset config

milo-hyben:
This PR adds 3 new attributes to dataset configuration schema:
display_name, team_ownership & billing_groups

### DIFF
--- a/cpg_infra/config/config.py
+++ b/cpg_infra/config/config.py
@@ -399,7 +399,9 @@ class CPGDatasetConfig(ConfigModel):
     display_name: str | None = None
 
     # Dataset's team ownership
-    team_ownership: str | None = None
+    team_ownership: Literal['Rare Disease', 'Population Genomics', 'Shared'] | None = (
+        None
+    )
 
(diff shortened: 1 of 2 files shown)


cpg_infra/config/config.py
```diff
@@ -395,6 +395,15 @@ class UploadDropboxConfig(ConfigModel):
     # creates a metamist project (+ test metamist project if setup_test is True)
     enable_metamist_project: bool = True
 
+    # Metamist Display name
+    display_name: str | None = None
+
+    # Dataset's team ownership
+    team_ownership: str | None = None
```
dancoates: Given we want to be strict on these I think this could be typed as a literal with the 3 available options. That'll ensure no typos slip through.

milo-hyben: Thank you Dan, very valid point. Had made a change.

violetbrina (approved):

dancoates (approved):

PR #381, opened 2026-08-10
SET-1247 Prepare cpg-infra repo for Seqera implementation

milo-hyben:
## PR Summary:

This PR refactors cpg_infra/driver.py into a driver package. Prep step for the Seqera work.

- No logic or public API changes; class bodies moved verbatim.
- GroupMember, Group, and GroupProvider are co-located in driver/groups.py (closely related, not used independently).
- MainUploadBucket and SampleMetadataAccessorMembership are inlined into dataset_cloud_infrastructure.py (only used there).
- Type-only imports resolved with TYPE_CHECKING to break the CPGInfrastructure ↔ CPGDatasetCloudInfrastructure type-hint cycle.
- `from cpg_infra.driver import ...` remains stable.

## Future work

**FW-1. Make `NAME_TO_INFRA_CLASS` lazy** — before the next `CloudInfraBase` subclass lands
*Priority: high. Best done in the very next PR, ideally the one that
introduces Seqera.*

**FW-2. Retire the backward-compat class-attribute aliases**
*Priority: low. Do when the mono-repo has settled after Seqera.*

Will create Jira tickets for the above.


cpg_infra/abstraction/azure.py
```diff
@@ -393,6 +393,8 @@ def add_member_to_bucket(
     @staticmethod
     def _get_principal_type(obj):
         # it's a 'cpg_infra.driver.CPGInfrastructure.GroupProvider.Group'
+        # The canonical path is now `cpg_infra.driver.group.Group`,
```
dancoates: Not sure I understand this comment, is it just referring to the comment above? If so then the path can be updated no problems

cpg_infra/driver/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+from cpg_infra.driver.group_member import GroupMember
+from cpg_infra.driver.group_provider import GroupProvider
+from cpg_infra.driver.infrastructure import CPGInfrastructure
+from cpg_infra.driver.main_upload_bucket import MainUploadBucket
+from cpg_infra.driver.sm_accessor_membership import (
+    SampleMetadataAccessorMembership,
+)
+
+
+def test():
```
dancoates: I think we can just delete this, I didn't even know it was here and it isn't used anywhere in the repo

cpg_infra/driver/groups.py
```diff
@@ -0,0 +1,99 @@
...
+    from cpg_infra.config import CloudName
+    from cpg_infra.driver.group_member import GroupMember
+
+
+class GroupProvider:
+    """Provider for managing groups + memberships"""
+
+    # Legacy nested-class access paths: `GroupProvider.Group` /
+    # `GroupProvider.Group.GroupMember`
+    Group = _Group
```
dancoates: I think I'd prefer to not keep this legacy access path, it should be pretty easy to update references

cpg_infra/driver/main_upload_bucket.py
```diff
@@ -0,0 +1,14 @@
...
+"""
+MainUploadBucket named tuple.
+"""
+
+from __future__ import annotations
+
+from typing import Any, NamedTuple
+
+
+class MainUploadBucket(NamedTuple):
```
dancoates: I don't think this is generic enough to need being in its own file. It is only used in one place so could be in that file

cpg_infra/driver/sm_accessor_membership.py
```diff
@@ -0,0 +1,14 @@
...
+"""
+SampleMetadataAccessorMembership named tuple.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Iterable, NamedTuple
+
+
+class SampleMetadataAccessorMembership(NamedTuple):
```
dancoates: This is also only used in the dataset infra file so I think would be fine to be in there

test/test_driver_package.py
```diff
@@ -0,0 +1,87 @@
+"""
+Smoke tests for the ``cpg_infra.driver`` package.
```
dancoates: These tests are useful to validate the refactor, but I don't think have much use once the refactor is complete so I think should be deleted before this PR is merged.

cpg_infra/driver/groups.py
```diff
@@ -0,0 +1,99 @@
...
+# ``GroupProvider.Group`` via the class attribute.
+from cpg_infra.driver.group import Group as _Group
+
+if TYPE_CHECKING:
+    from cpg_infra.abstraction.base import CloudInfraBase
+    from cpg_infra.config import CloudName
+    from cpg_infra.driver.group_member import GroupMember
+
+
+class GroupProvider:
```
dancoates, line 97: My preference would be to keep the 3 group data structures in the one file seeing as they are closely related and not used separately - I don't feel super strongly though if you'd prefer to keep them in their own files.

dancoates (commented):
This looks good, just a few small comments around the structure.

I've also tested this with a script that reconstructs the old `driver.py` from the files it has been split into so that I could diff that against the `driver.py` on the main branch to confirm that this PR contains no logic changes - and can confirm that there are no changes other than shuffling files around. 




<details>

<summary>Here's the script if anyone else who happens to be looking at this wants to test:</summary>



```python
#!/usr/bin/env python3
from __future__ import annotations

import ast
from pathlib import Path

DRIVER_DIR = Path(__file__).parent / 'cpg_infra' / 'driver'
OUT = Path(__file__).parent / 'driver_merged.py'

# Concatenation order chosen to mirror the original driver.py definition order.
ORDER = [
    'sm_accessor_membership.py',
    'constants.py',
    'main_upload_bucket.py',
    'group_member.py',
    'group.py',
    'group_provider.py',
    'infrastructure.py',
    'dataset_infrastructure.py',
    'dataset_cloud_infrastructure.py',
    '__init__.py',
]


def strip_preamble(source: str) -> str:
    """Return the file body with its leading module docstring, __future__/
    imports and top-level ``if TYPE_CHECKING:`` guards removed.

    Slices from the first "real" top-level statement (class/def/assignment) to
    the end of the file, so anything after that first definition is preserved
    verbatim.
    """
    tree = ast.parse(source)
    lines = source.splitlines()

    for node in tree.body:
        # module docstring
        if (
            isinstance(node, ast.Expr)
            and isinstance(node.value, ast.Constant)
            and isinstance(node.value.value, str)
        ):
            continue
        # plain imports / from ... import ...
        if isinstance(node, (ast.Import, ast.ImportFrom)):
            continue
        # top-level `if TYPE_CHECKING:` guard blocks
        if isinstance(node, ast.If):
            test = node.test
            if isinstance(test, ast.Name) and test.id == 'TYPE_CHECKING':
                continue
        # first real definition -> keep from here to EOF
        return '\n'.join(lines[node.lineno - 1:]).rstrip() + '\n'

    return ''


def main() -> None:
    chunks: list[str] = []
    for name in ORDER:
        path = DRIVER_DIR / name
        body = strip_preamble(path.read_text())
        if not body.strip():
            continue
        chunks.append(f'# ===== from {name} =====\n{body}')

    OUT.write_text('\n\n'.join(chunks) + '\n')
    print(f'Wrote {OUT} ({OUT.read_text().count(chr(10))} lines)')


if __name__ == '__main__':
    main()

```

</details>

cpg_infra/abstraction/azure.py
milo-hyben: Comment updated, so hopefully it makes more sense.

cpg_infra/driver/__init__.py
milo-hyben: Removed

cpg_infra/driver/groups.py
milo-hyben: Legacy access path removed, references updated.

cpg_infra/driver/main_upload_bucket.py
milo-hyben: Make sense, moved to dataset_cloud)infrastracture.py

cpg_infra/driver/sm_accessor_membership.py
milo-hyben: The same, moved to the dataset infra file.

test/test_driver_package.py
milo-hyben: test file removed

cpg_infra/driver/groups.py
milo-hyben: All 3 classes are now in one groups.py file.

dancoates (approved):
Thanks for the updates, this is looking good - I think just needs to fix linting/formatting and it'd be good to see a preview run on infra-private to confirm that this makes no changes to the stack.

milo-hyben:
Yes, thanks Dan, will be testing preview shortly.

dancoates (approved):

PR #382, opened 2026-08-11
Fix for meta update for metamist projects.

milo-hyben:
Resolve project_id defensively:
- new state stores it under 'project_id' (commit 295359f onwards)
- state from before that rename stored it under 'id'
- if neither is present (e.g. stale refresh), re-fetch by name


violetbrina (approved):
LGTM

PR #383, opened 2026-08-11
Relaxing type for Pulumi's runtime.

milo-hyben:
This PR make changes to fix the pulumi runtime errors:

- Class annotation meta relaxed from pulumi.Output[dict[str, str | list[str]]] to pulumi.Output[dict[str, Any]], this is the one Pulumi's runtime introspects

- Constructor parameter meta: kept as dict[str, str | list[str]] | None — this contract is only read by callers and type checkers, so the tight typing stays.

- Added a short comment explaining why the runtime annotation has to be wider than the input contract, so nobody tightens it back later without hitting the same crash.


violetbrina (approved):
LGTM

PR #387, opened 2026-08-19
[SET-1211] Create Seqera Resources for the first set of projects

rameshka:
WIP

- For each dataset, create new GCS buckets as work directories for compute environments in `main` and `test` workspaces.
- In the current implementation
  - The compute environment created for each dataset has the same `compute` config
  - The service account created for the WIF is used as the compute environment's runtime SA.

**TODO**: Add description


cpg_infra/driver/dataset_seqera_infrastructure.py
```diff
@@ -153,20 +204,164 @@ def _grant_project_roles(self) -> None:
...
+        # reference to participants in workspaces
+        # the same member can be in different datasets (under same or different teamOwnership)
+        # but they will be added once to the respective workspace
+        workspace_participants: set[tuple[TeamOwnership, str, MemberKey]] = (
+            root.seqera_workspace_participants
+        )
+        gcp_key = GcpInfrastructure.name()
+        team_name = _format_team_name(team)
+
+        members_to_add: list[MemberKey] = self._dataset_config.members.get(
```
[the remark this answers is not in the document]
rameshka: Addressed in https://github.com/populationgenomics/cpg-infrastructure/pull/387/commits/93db22aa927d225e4821fcf220d750ba0e461506

cpg_infra/driver/dataset_seqera_infrastructure.py
```diff
@@ -7,7 +7,7 @@
 
 Two service accounts are created per access level:
   - Head Job SA (seqera-{level}-head): launches batch jobs; no dataset access.
-  - Task Job SA (seqera-{level}): executes tasks with dataset access; no launch permission.
+  - Task Job SA (seqera-{level}): executes tasks with dataset access.
```
milo-hyben:
```suggestion
  - Task Job SA (seqera-{level}): executes tasks with dataset access  and can spawn nested jobs.
  
Separation of concerns:
  - Head SA: Launches the primary workflow
  - Task SA: Executes tasks and can launch nested subtasks (e.g., Dataproc jobs)
```

cpg_infra/driver/dataset_seqera_infrastructure.py
```diff
@@ -37,12 +53,12 @@
...
     'roles/batch.jobsEditor',
     'roles/logging.viewer',
-    'roles/iam.workloadIdentityUser',
 )
 
-# Task Job SA runs the actual compute tasks; no job-launch permission.
+# Task Job SA runs the actual compute tasks.
 _TASK_JOB_ROLES: tuple[str, ...] = (
     'roles/batch.agentReporter',
+    'roles/batch.jobsEditor',
```
milo-hyben:
```suggestion
    'roles/batch.jobsEditor',  # Allows spawning nested jobs (e.g., Dataproc)
```

cpg_infra/driver/dataset_seqera_infrastructure.py
```diff
@@ -37,12 +53,12 @@
 _HEAD_JOB_ROLES: tuple[str, ...] = (
     'roles/batch.jobsEditor',
     'roles/logging.viewer',
-    'roles/iam.workloadIdentityUser',
 )
 
-# Task Job SA runs the actual compute tasks; no job-launch permission.
+# Task Job SA runs the actual compute tasks.
```
milo-hyben:
```suggestion
# Task Job SA runs the actual compute tasks and can spawn nested subtasks
# (e.g., Dataproc jobs within a Nextflow task). Head SA launches the main workflow;
# Task SA executes tasks and can launch nested jobs for those tasks.
```

milo-hyben (commented):
Looking very good Piyumi! 
I only have few small suggestions to class names and extended few comments.

cpg_infra/config/config.py
```diff
@@ -28,6 +28,22 @@
     'external-repository-reader',
 ]
 
+TeamOwnership = Literal['Rare Disease', 'Population Genomics', 'Shared']
+
+
+class SeqeraWorkspaceRef(ConfigModel):
```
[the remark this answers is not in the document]
rameshka: I have another class, `SeqeraWorkspace`, in the dynamic provider. I was using the `Ref` suffix for distinuguish the instances created from the incoming configs. (It's not a good choice for naming.)

`SeqeraWorkspaceRef` and `SeqeraWorkspaceRefPair` are no longer needed now as I'm reusing the class definitions that you have newly declared.

cpg_infra/driver/dataset_seqera_infrastructure.py
```diff
@@ -37,12 +57,14 @@
...
-    'roles/iam.workloadIdentityUser',
 )
 
-# Task Job SA runs the actual compute tasks; no job-launch permission.
+# Task Job SA runs the actual compute tasks and can spawn nested subtasks
+# (e.g., Dataproc jobs within a Nextflow task). Head SA launches the main workflow;
+# Task SA executes tasks and can launch nested jobs for those tasks.
 _TASK_JOB_ROLES: tuple[str, ...] = (
     'roles/batch.agentReporter',
+    'roles/batch.jobsEditor',  # Allows spawning nested jobs (e.g., Dataproc)
```
dancoates: I'm not sure if this is correct about allowing Dataproc jobs. Dataproc doesn't run on Batch, it creates its own Dataproc cluster and I think would need a different set of permissions?

dancoates (dismissed):
This is looking really good @rameshka, nice work! I just had one small note on the task job roles. I don't think we necessarily need to nail dataproc permissions in this PR but I'm not sure the current comment is accurate in respect to dataproc. 

Only other small thing is to make sure the todos are removed from the commit message that is generated from the PR description. 

And would be good to have approval from @milo-hyben as well before merging, thanks!

cpg_infra/driver/dataset_seqera_infrastructure.py
rameshka: Oh, I will update the comment, @dancoates. I tested Batch jobs, and they require this permission to spawn child jobs. I assumed the same for Dataproc.

rameshka:
Thank you very much for reviewing @dancoates  and @milo-hyben. 

I will make sure the final commit message is proper.

cpg_infra/driver/dataset_seqera_infrastructure.py
rameshka: addressed in https://github.com/populationgenomics/cpg-infrastructure/pull/387/commits/264950779adbe5e0fba27661e65a4adb32569355

milo-hyben (approved):
Looking good @rameshka, nice work!

milo-hyben (approved):
LGTM

